PR #16, opened 2020-04-17
Update to Rust nightly-2020-03-22

spernsteiner:
Aside from the usual updates to account for rustc refactoring, this also required a bunch of improvements to constant handling.  See GaloisInc/mir-verifier#23 for more info.

### DIFF
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -561,83 +561,6 @@ fn mir_body(
 }
 
 
-/*
-/// Workaround for a rustc bug.  MIR bodies are usually stored in normalized form, but for
-/// associated constants whose types involve parameters, one occurrence of the constant type is not
-/// substituted or normalized properly:
-///
-///     trait TheTrait {
-///         const THE_CONST: Self;
-///     }
-///
-///     impl TheTrait for u32 {
-///         const THE_CONST: u32 = 123;
-///     }
-///
-///
-///     mir::Constant {
-///         ty: u32,        // Substituted + normalized type of <u32 as TheTrait>::THE_CONST
-///         literal: Const {
-///             ty: Self,   // Unsubstituted declared type of THE_CONST.  This is the bug.
-///             val: Unevaluated(
-///                 the_crate::TheTrait::THE_CONST,
-///                 [u32],  // Substs for TheTrait::THE_CONST
-///             ),
-///         },
-///     }
-///
-/// This pass looks for Unevaluated constants, and copies the outer ty (which is computed
-/// correctly) over top of the inner one.  Afterward, the MIR is well-formed and running `subst` on
-/// it no longer panics.
-fn subst_const_tys<'tcx>(
-    tcx: TyCtxt<'tcx>,
-    substs: ty::subst::SubstsRef<'tcx>,
-    body: &mut Body<'tcx>,
-) {
-    mir::visit::MutVisitor::visit_body(
-        &mut SubstConstTys { tcx, substs },
-        body,
-    );
-    for promoted in body.promoted.iter_mut() {
-        subst_const_tys(tcx, substs, promoted);
-    }
-}
-
-struct SubstConstTys<'tcx> {
-    tcx: TyCtxt<'tcx>,
-    substs: ty::subst::SubstsRef<'tcx>,
-}
-
-impl<'tcx> mir::visit::MutVisitor<'tcx> for SubstConstTys<'tcx> {
-    fn tcx(&self) -> TyCtxt<'tcx> {
-        self.tcx
-    }
-
-    fn visit_constant(
-        &mut self,
-        constant: &mut mir::Constant<'tcx>,
-        location: mir::Location,
-    ) {
-        if let mir::interpret::ConstValue::Unevaluated(..) = constant.literal.val {
-            constant.literal = self.tcx.mk_const(ty::Const {
-                ty: constant.ty,
-                val: constant.literal.val,
-            });
-        }
-        self.super_constant(constant, location);
-    }
-}
-*/
-fn subst_const_tys<'tcx>(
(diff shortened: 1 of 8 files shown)


src/analyz/mod.rs
```diff
@@ -61,11 +65,12 @@ impl<'tcx> ToJson<'tcx> for mir::AggregateKind<'tcx> {
             &mir::AggregateKind::Adt(_, _, _, _, _) => {
                 panic!("adt should be handled upstream")
             }
-            &mir::AggregateKind::Closure(ref defid, ref closuresubsts) => {
+            &mir::AggregateKind::Closure(ref defid, ref substs) => {
                 json!({
                     "kind": "Closure",
                     "defid": defid.to_json(mir),
-                    "closuresubsts": closuresubsts.substs.to_json(mir)
+                    // FIXME rename
```
atomb, line 72: Is this left unchanged because it'll require changes to the `mir-verifier` parser, too?

src/analyz/mod.rs
```diff
@@ -292,8 +273,14 @@ impl<'tcx> ToJson<'tcx> for mir::PlaceElem<'tcx> {
...
             }
-            &mir::ProjectionElem::Subslice { ref from, ref to } => {
-                json!({"kind": "Subslice", "from": from, "to": to})
+            &mir::ProjectionElem::Subslice { ref from, ref to, from_end } => {
+                json!({
+                    "kind": "Subslice",
+                    "from": from,
+                    "to": to,
+                    // FIXME add to mir-verifier
+                    "from_end": from_end,
```
atomb, line 282: Has this been added yet?

src/analyz/mod.rs
```diff
@@ -331,7 +318,8 @@ impl<'tcx> ToJson<'tcx> for mir::Operand<'tcx> {
 impl<'tcx> ToJson<'tcx> for mir::Constant<'tcx> {
     fn to_json(&self, mir: &mut MirState<'_, 'tcx>) -> serde_json::Value {
         json!({
-            "ty": self.ty.to_json(mir),
+            // FIXME remove
+            "ty": self.literal.ty.to_json(mir),
```
atomb, line 322: Are types no longer parts of constants? Or are they part of the literal field?

src/analyz/mod.rs
```diff
@@ -621,6 +610,10 @@ struct SubstConstTys<'tcx> {
 }
 
 impl<'tcx> mir::visit::MutVisitor<'tcx> for SubstConstTys<'tcx> {
+    fn tcx(&self) -> TyCtxt<'tcx> {
+        self.tcx
+    }
+
```
atomb, line 616: This function seems added and commented out. Is it going to be used eventually?

src/analyz/mod.rs
```diff
@@ -635,6 +628,14 @@ impl<'tcx> mir::visit::MutVisitor<'tcx> for SubstConstTys<'tcx> {
...
     }
 }
+*/
+fn subst_const_tys<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    substs: ty::subst::SubstsRef<'tcx>,
+    body: &mut Body<'tcx>,
+) {
+    // FIXME remove
+}
```
atomb, line 638: Can this be removed already? It seems like it's called once, for side effects, and it doesn't have any.

src/analyz/mod.rs
```diff
@@ -959,7 +980,8 @@ fn emit_fn<'tcx>(
         "generics": { "params": [] },
         "predicates": { "predicates": [] },
         "body": mir_body(ms),
-        "promoted": promoted,
+        // FIXME remove
+        "promoted": [],
```
atomb, line 984: Can this safely be removed yet? Or will the parser break?

src/analyz/ty_json.rs
```diff
@@ -857,44 +928,285 @@ fn render_constant<'tcx>(
...
+        ) -> InterpResult<'tcx, u64> {
+            Err(InterpError::Unsupported(
+                UnsupportedOpInfo::Unsupported(
+                    "ptr_to_int".into(),
+                ),
+            ).into())
+        }
+    }
+}
+
```
atomb, line 1179: Wow. Constants are complicated!

src/analyz/mod.rs
spernsteiner: It actually has - I'll remove the comment.

src/analyz/mod.rs
spernsteiner: Yes, the `literal` field contains the `ty` now.

I looked at the mir-verifier parser for this case, and it looks like it could use some significant cleanup.  It handles three different `Literal` variants, but the current version of mir-json only ever emits one.  I'll make a separate branch for JSON cleanup after this, to cover `Literal` and the other JSON format FIXMEs you noted.

src/analyz/mod.rs
spernsteiner: This was part of an incomplete update of `subst_const_tys` for the new Rust nightly, but `subst_const_tys` can just be removed instead.

src/analyz/mod.rs
spernsteiner: Well, all the tests are passing, so I guess it's safe to remove.  `subst_const_tys` was a workaround for a rustc bug, which seems to be fixed in the new nightly.

src/analyz/ty_json.rs
spernsteiner: Yep!  Fortunately, we were able to use rustc's const evaluation infrastructure (with this dummy `Machine` impl) to read out the discriminants and fields of evaluated constants.  Someday we may need to implement that functionality ourselves to handle transmutes in mir-verifier, but for now, this saves us a lot of work.

PR #20, opened 2020-10-05
Add a wrapper that sets the library path

spernsteiner:
This renames every binary `foo` to `foo_real`, and adds a wrapper `foo` that sets `$(DY)LD_LIBRARY_PATH` appropriately before invoking `foo_real`.  The library path is determined automatically by invoking `rustc +nightly-2020-03-22 --print sysroot`, so it should come up with the right path regardless of `rustup` install location.

Tested on Linux and macOS

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,6 +22,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.32"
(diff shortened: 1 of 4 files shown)


atomb (approved):
Awesome! Thanks, @spernsteiner!

Issue #43, opened 2023-04-24
Rename `CRUX_RUST_LIBRARY_PATH` to `RUST_LIBRARY_PATH`


RyanGlScott:
The `CRUX_RUST_LIBRARY_PATH` is a very important environment variable used to locate the Rust standard library files that `mir-json` needs to be able to link MIR JSON files properly. The `CRUX_` part of the name is somewhat outdated in lieu of [`mir-json` being able to support SAW](https://github.com/GaloisInc/mir-json/pull/41), however, as Crux is no longer the only tool that `mir-json` supports. We should rename `CRUX_RUST_LIBRARY_PATH` to `RUST_LIBRARY_PATH` to reflect this more general functionality.

For the sake of backwards compatibility, it might be nice to keep `CRUX_RUST_LIBRARY_PATH` around for a while as a deprecated alias for `RUST_LIBRARY_PATH`.

PR #46, opened 2023-05-03
Add `SAW_RUST_LIBRARY_PATH` alias for `CRUX_RUST_LIBRARY_PATH`

RyanGlScott:
Fixes https://github.com/GaloisInc/mir-json/issues/43.

### DIFF
--- a/doc/rustc.md
+++ b/doc/rustc.md
@@ -11,8 +11,11 @@
   invoke normal `rustc` compilation with some additional callbacks installed.
 
 These tools require a version of the Rust standard library that has been
-translated with `mir-json`. The `CRUX_RUST_LIBRARY_PATH` environment variable
-should be defined with this path before using one of these tools.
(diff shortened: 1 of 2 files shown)


RyanGlScott:
I'd be fine with `CRUCIBLE_RUST_LIBRARY_PATH` and keeping the rest unchanged. I'll push a new version which does this.

RyanGlScott:
Hm. Even if we rename it to `CRUCIBLE_RUST_LIBRARY_PATH`, there is an additional awkward tension in that all of the relevant `.rlib`/`.mir` files currently live in `crux-mir`. I suppose that when we distribute SAW in the future, we will want to include these `.rlib`/`.mir` files in the SAW bindist so that they can easily be located.

You know what? I think you've convinced me to pick `SAW_RUST_LIBRARY_PATH`, as I think that name will be less confusing for people who don't know what Crucible is.

spernsteiner (approved):

Issue #47, opened 2023-06-29
Unexpected `saw-rustc` compilation failure (`#[test] can only be applied to functions`)


RyanGlScott:
`saw-rustc` fails to compile this rather simple Rust program:

```rs
pub fn f() -> i32 {
    let xs : [i32; 3] = [0; 3];
    xs[0]
}
```
```
$ SAW_RUST_LIBRARY_PATH=~/Documents/Hacking/Haskell/crucible/crux-mir/rlibs saw-rustc test.rs 
test build - extract output path - ["rustc", "test.rs", "--test", "--target", "x86_64-unknown-linux-gnu", "--cfg", "crux", "-L", "/home/rscott/Documents/Hacking/Haskell/crucible/crux-mir/rlibs"]
test build - ["rustc", "test.rs", "--target", "x86_64-unknown-linux-gnu", "--cfg", "crux", "-L", "/home/rscott/Documents/Hacking/Haskell/crucible/crux-mir/rlibs", "--cfg", "crux_top_level", "--crate-type", "rlib"]
error: #[test] can only be applied to functions
 --> test.rs:2:29
  |
2 |     let xs : [i32; 3] = [0; 3];
  |                             ^

error: #[test] can only be applied to functions
 --> test.rs:2:20
  |
2 |     let xs : [i32; 3] = [0; 3];
  |                    ^

note: Emitting MIR for test/3a1fbbbh::f[0]

error: aborting due to 2 previous errors

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ErrorReported', src/bin/mir-json-rustc-wrapper.rs:255:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I'm not entirely sure what is going on here, but my hunch is that `saw-rustc` is inserting `#[test]` attributes where it shouldn't.

Issue #32, opened 2023-01-13
Needs updated for split of rustc_mir module into many smaller modules


TomMD:
See https://github.com/rust-lang/rust/pull/80522

There is no more rustc-mir and that change will require lots of changes here.

Issue #31, opened 2022-10-19
README compile instructions not working


robdockins:
Maybe due to LLVM version drift? Any chance we could get some binary builds?

```
% rustc --version
rustc 1.44.0-nightly (38114ff16 2020-03-21)

% uname -a
Darwin <snip> 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:17:10 PDT 2022; root:xnu-8020.140.49~2/RELEASE_X86_64 x86_64

 % clang --version
Apple clang version 14.0.0 (clang-1400.0.29.102)
Target: x86_64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

% git log | head -n 5
commit fbacd2a8d5885a3897c7b3efe4b61607aed9cab2
Author: Stuart Pernsteiner <spernsteiner@galois.com>
Date:   Thu Apr 7 15:33:38 2022 -0700

    allow changing crux-mir binary path using CRUX_MIR env var

% cargo install --locked
warning: Using `cargo install` to install the binaries for the package in current working directory is deprecated, use `cargo install --path .` instead. Use `cargo build` if you want to simply build the package.
  Installing mir-json v0.1.0 (/Users/rdoc/code/mir-json)
    Updating crates.io index
warning: package `miow v0.3.3` in Cargo.lock is yanked in registry `crates.io`, consider running without --locked
warning: package `socket2 v0.3.11` in Cargo.lock is yanked in registry `crates.io`, consider running without --locked
   Compiling libc v0.2.65
   Compiling proc-macro2 v1.0.7
   Compiling syn v1.0.13
   Compiling serde v1.0.104
   Compiling memchr v2.3.0
   Compiling getrandom v0.1.14
   Compiling log v0.4.8
   Compiling bitflags v1.2.1
   Compiling ryu v0.2.7
   Compiling typenum v1.11.2
   Compiling regex-syntax v0.6.13
   Compiling percent-encoding v2.1.0
error: linking with `cc` failed: exit code: 1
  |
[...]

Issue #22, opened 2022-06-02
Arm64 Support (M1)


franziskuskiefer:
The `nightly-2020-03-22-aarch64-apple-darwin` toolchain used by mir-json right now does not exist.
Are there plans to bump the nightly version to something a little newer?

bhargavbh:
I have faced similar issue. Would be great if nightly version is upgraded to something that has `aarch64-apple-darwin` target

PR #49, opened 2023-07-03
Support `nightly-2023-01-23`

RyanGlScott:
This is a fairly large overhaul of `mir-json`'s internals to support a more recent Rust nightly (`nightly-2023-01-23`). The work is split between myself, @m10f, and @spernsteiner.

There is a lot happening here, so if you want to learn specifics, I'd encourage you to look at individual commits. Some of the broad themes in the patchset include:

* Propertly support for Rust's new constant forms
* Better support for zero-sized constants
* Encoding `enum` discriminant types so that `crucible-mir` can know about non-`isize` discriminant types (e.g., `Ordering`, which uses an `i8` discriminant)
* Make `mir-json` emit a `--sysroot` argument when invoking `rustc` so that `rustc` knows how to disambiguate it from the default sysroot
* Adding a special case for `&str`, analogous to a special case for `&[T]`
* Ensure that `AnonConst`s aren't marked with `#[test]`

This should fix several issues:

* Fixes https://github.com/GaloisInc/mir-json/issues/22
* Fixes https://github.com/GaloisInc/mir-json/issues/31
* Fixes https://github.com/GaloisInc/mir-json/issues/32
* Fixes https://github.com/GaloisInc/mir-json/issues/47


src/analyz/mod.rs
```diff
@@ -658,14 +725,50 @@ fn emit_static_decl<'tcx>(
...
 fn has_test_attr(tcx: TyCtxt, def_id: DefId) -> bool {
-    def_id.is_local() && tcx.has_attr(def_id, Symbol::intern("crux_test"))
+    if !def_id.is_local() {
+        return false;
+    }
+
+    for attr in tcx.get_attrs_unchecked(def_id) {
+        let crux = Symbol::intern("crux");
+        let test = Symbol::intern("test");
+        match &attr.kind {
```
spernsteiner, line 743:
```suggestion
    let crux = Symbol::intern("crux");
    let test = Symbol::intern("test");
    for attr in tcx.get_attrs_unchecked(def_id) {
        match &attr.kind {
```
No sense re-interning the same constant strings with every iteration

src/analyz/mod.rs
```diff
@@ -658,14 +725,50 @@ fn emit_static_decl<'tcx>(
...
+
+    // writeln!(file, "has crux::test attr:{}", tcx.has_attr(def_id, Symbol::intern("crux::test"))).unwrap();
+    // for a in attrs {
+    //     writeln!(file, "attr: {:?}", a).unwrap();
+    // }
+    // file.flush().unwrap();
+
+    // for
+
+    // def_id.is_local() && tcx.has_attr(def_id, Symbol::intern("crux::test"))
```
spernsteiner, line 771:
```suggestion
```
Looks like commented-out debug code

src/analyz/mod.rs
```diff
@@ -1087,29 +1205,36 @@ pub use analyz::to_json::ExportStyle;
...
         span: Span::default(),
     }
 }
 
 pub fn inject_attrs<'tcx>(queries: &'tcx Queries<'tcx>) {
-    let mut krate = queries.parse().unwrap().peek_mut();
-    krate.attrs.push(make_attr("feature", "register_attr"));
-    krate.attrs.push(make_attr("register_attr", "crux_test"));
+    let mut k = queries.parse().unwrap(); // need to call `get_mut`?
+    let krate: &mut Crate = k.get_mut();
```
spernsteiner, line 1235:
```suggestion
    let mut k = queries.parse().unwrap();
    let krate: &mut Crate = k.get_mut();
```

src/analyz/mod.rs
```diff
@@ -1177,10 +1302,17 @@ thread_local! {
...
+        let resolver = tcx.resolver_for_lowering(());
+        let krate = &resolver.borrow().1;
+        let mut v = GatherMatchSpans::default();
+        visit::walk_crate(&mut v, krate);
+        MATCH_SPAN_MAP.with(|m| m.replace(Some(Rc::new(v.match_span_map))));
+    });
+    // let k = queries.expansion().unwrap();
+    // let krate: &ast::Crate = &k.borrow().0;
+
+
```
spernsteiner, line 1315:
```suggestion
```

src/analyz/to_json.rs
```diff
@@ -107,22 +108,43 @@ impl<'tcx> TraitInst<'tcx> {
...
+                    ty::AssocKind::Type => {},
+                    _ => continue,
+                }
+                let proj_ty = tcx.mk_projection(ai.def_id, trait_ref.substs);
+                let actual_ty = tcx.normalize_erasing_regions(ty::ParamEnv::reveal_all(), proj_ty);
+                projs.push(ty::ExistentialProjection {
+                    def_id: ai.def_id,
+                    substs: ex_trait_ref.substs,
+                    term: actual_ty.into(), // TO REVIEW: structure of Term has changed
+                });
```
spernsteiner, line 144: What does this "to review" comment refer to?

src/analyz/to_json.rs
```diff
@@ -133,10 +155,13 @@ impl<'tcx> TraitInst<'tcx> {
...
-        preds.extend(self.projs.iter().map(|p| ty::ExistentialPredicate::Projection(*p)));
-        let preds = tcx.intern_existential_predicates(&preds);
-        Some(tcx.mk_dynamic(ty::Binder::bind(preds), tcx.mk_region(ty::RegionKind::ReErased)))
+        preds.push(ty::Binder::dummy(ty::ExistentialPredicate::Trait(trait_ref)));
+        preds.extend(
+            self.projs.iter().map(|p| ty::Binder::dummy(ty::ExistentialPredicate::Projection(*p))),
+        );
+        let preds = tcx.intern_poly_existential_predicates(&preds);
+        // TO REVIEW: DynKind?
+        Some(tcx.mk_dynamic(preds, tcx.mk_region(ty::RegionKind::ReErased), DynKind::Dyn))
```
spernsteiner, line 164: `DynKind` exists to distinguish normal `dyn Trait` from the experimental `dyn* Trait` type.  I don't think anybody really uses `dyn*` at the moment, so it's okay if we don't support it.

```suggestion
        // Always emit `DynKind::Dyn`.  We don't support `dyn*` (`DynKind::DynStar`) yet.
        Some(tcx.mk_dynamic(preds, tcx.mk_region(ty::RegionKind::ReErased), DynKind::Dyn))
```

src/analyz/ty_json.rs
```diff
@@ -412,35 +419,35 @@ impl<'tcx> ToJson<'tcx> for ty::Ty<'tcx> {
...
+                    "upvar_tys": substs.as_closure().upvar_tys()
                         .collect::<Vec<_>>().to_json(mir),
                     // crucible-mir uses the same representation for closures as it does for
                     // tuples, so no additional information is needed.
                 })
             }
-            &ty::TyKind::Dynamic(ref preds, _region) => {
-                let ti = TraitInst::from_dynamic_predicates(mir.state.tcx, *preds);
+            &ty::TyKind::Dynamic(preds, _region, _dynkind) => {
+                let ti = TraitInst::from_dynamic_predicates(mir.state.tcx, preds);
```
spernsteiner, line 429: Maybe we should require the last field here to be `DynKind::Dyn` (and have a separate case similar to `TyKind::Error` for `DynKind::DynStar`).  It's better for code using `dyn*` to fail explicitly rather than trigger some weird interaction between this code and the `TraitInst` code above that only handles the `DynKind::Dyn` case.

src/analyz/ty_json.rs
```diff
@@ -662,329 +658,108 @@ impl<'tcx> ToJson<'tcx> for ty::subst::GenericArg<'tcx> {
...
-        },
-        _ => panic!("impossible: array size is not a scalar?"),
+    match c.kind() {
+        // ty::ConstKind::Unevaluated(un) => {
+        //     tcx.const_eval_resolve(ty::ParamEnv::reveal_all(), un.expand(), None).unwrap()
+        // },
+        // NB: the type of "val" changed to valtree - try to get the size out of that
+        ty::ConstKind::Value(ty::ValTree::Leaf(val)) =>
+            val.try_to_machine_usize(tcx).expect("expecting usize value from constant") as usize,
+        ref val => panic!("don't know how to translate ConstKind::{:?}", val),
```
spernsteiner, line 672: Could you add some tests in crux-mir that try to exercise the `Unevaluated` case?  It looks like we don't have any at the moment.

```Rust
const FOO: usize = 5;

fn f(xs: [i32; FOO]) -> i32 {
    xs[2]
}


trait HasSize {
    const SIZE: usize;
}

impl HasSize for () {
    const SIZE: usize = 6;
}

fn g<T: HasSize>(xs: [i32; T::SIZE]) -> i32 {
    xs[2]
}

// Do something with `f` and `g::<()>`...
```

If these pass, then the commented-out case can be safely removed.  (And if they don't, we can mark them as expected-fail / known bug and remove the case anyway.)

src/analyz/ty_json.rs
```diff
@@ -662,329 +658,108 @@ impl<'tcx> ToJson<'tcx> for ty::subst::GenericArg<'tcx> {
...
     }
+    // match evaluated {
+    //     interpret::ConstValue::Scalar(interpret::Scalar::Int(sint)) => {
+    //         let data = sint.to_bits(sint.size()).unwrap();
+    //         assert!(data <= usize::MAX as u128);
+    //         data as usize
+    //     },
+    //     _ => panic!("impossible: array size is not a scalar?"),
+    // }
```
spernsteiner, line 681:
```suggestion
```

src/analyz/ty_json.rs
```diff
@@ -662,329 +658,108 @@ impl<'tcx> ToJson<'tcx> for ty::subst::GenericArg<'tcx> {
...
+        }
+
+        // fn enforce_number_init(ecx: &InterpCx<'mir, 'tcx, Self>) -> bool {
+        //     false
+        // }
+
+        // fn enforce_number_no_provenance(ecx: &InterpCx<'mir, 'tcx, Self>) -> bool {
+        //     false
+        // }
+
```
spernsteiner, line 753:
```suggestion
```

src/analyz/ty_json.rs
```diff
@@ -1135,6 +936,408 @@ impl<'tcx> ToJson<'tcx> for ty::Const<'tcx> {
...
+    }
+}
+
+// fn eval_op<'tcx>(op_ty: interpret::OpTy<'tcx>) -> interpret::Scalar<interpret::AllocId> {
+//     match *op_ty {
+//         interpret::Operand::Immediate(imm) => imm.to_scalar(),
+//         interpret::Operand::Indirect(place) => place.,
+//     }
+// }
+
```
spernsteiner, line 967:
```suggestion
```

src/analyz/ty_json.rs
```diff
@@ -1135,6 +936,408 @@ impl<'tcx> ToJson<'tcx> for ty::Const<'tcx> {
...
+                let len = mplace_ty_len(&d, icx).unwrap();
+                let mem = icx.read_bytes_ptr_strip_provenance(d.ptr, Size::from_bytes(len)).unwrap();
+                return Some(json!({
+                    "kind": "str",
+                    "val": mem
+                }))
+            },
+            // Special case for &u8
+            ty::TyKind::Array(elem_ty, _) => {
+                if let ty::TyKind::Uint(ty::UintTy::U8) = *elem_ty.kind() {
```
spernsteiner, line 1203:
```suggestion
            // Special case for &[u8; N]
            ty::TyKind::Array(elem_ty, _) => {
                if let ty::TyKind::Uint(ty::UintTy::U8) = *elem_ty.kind() {
```

src/analyz/ty_json.rs
```diff
@@ -1135,6 +936,408 @@ impl<'tcx> ToJson<'tcx> for ty::Const<'tcx> {
...
+
+// TODO: move
+pub fn mk_interp<'mir, 'tcx>(tcx: TyCtxt<'tcx>) -> interpret::InterpCx<'mir, 'tcx, RenderConstMachine<'mir, 'tcx>> {
+    interpret::InterpCx::new(
+        tcx,
+        DUMMY_SP,
+        ty::ParamEnv::reveal_all(),
+        RenderConstMachine::new(),
+    )
+}
```
spernsteiner, line 1301: Seems this function is only called in one place, so the logic could be inlined there.

src/analyz/ty_json.rs
```diff
@@ -1262,3 +1512,50 @@ pub fn handle_adt_ag<'tcx>(
...
+    //         ErrorHandled::TooGeneric => {
+    //             self.cx
+    //                 .tcx()
+    //                 .sess
+    //                 .diagnostic()
+    //                 .emit_bug(errors::PolymorphicConstantTooGeneric { span: constant.span });
+    //         }
+    //     }
+    //     err
+    // }).unwrap()
```
spernsteiner, line 1560:
```suggestion
    tcx.const_eval_resolve(ty::ParamEnv::reveal_all(), uv, None).unwrap()
```

src/analyz/ty_json.rs
```diff
@@ -1262,3 +1512,50 @@ pub fn handle_adt_ag<'tcx>(
...
+) -> interpret::OpTy<'tcx> {
+    let layout =  tcx.layout_of(ty::ParamEnv::reveal_all().and(constant.ty())).unwrap();
+    icx.eval_mir_constant(&constant.literal, Some(constant.span), Some(layout)).unwrap()
+}
+
+pub fn eval_mir_constant<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    constant: &mir::Constant<'tcx>,
+) -> interpret::ConstValue<'tcx> {
+    // let ct = self.monomorphize(constant.literal);
```
spernsteiner, line 1529:
```suggestion
```

src/analyz/ty_json.rs
```diff
@@ -1262,3 +1512,50 @@ pub fn handle_adt_ag<'tcx>(
...
+    let uv = match constant.literal {
+        mir::ConstantKind::Ty(ct) => match ct.kind() {
+            ty::ConstKind::Unevaluated(uv) => uv.expand(),
+            ty::ConstKind::Value(val) => {
+                return tcx.valtree_to_const_val((ct.ty(), val));
+            }
+            err => panic!(
+                // constant.span,
```
spernsteiner, line 1537:
```suggestion
```

src/analyz/ty_json.rs
```diff
@@ -1262,3 +1512,50 @@ pub fn handle_adt_ag<'tcx>(
...
+    constant: &mir::Constant<'tcx>,
+) -> interpret::OpTy<'tcx> {
+    let layout =  tcx.layout_of(ty::ParamEnv::reveal_all().and(constant.ty())).unwrap();
+    icx.eval_mir_constant(&constant.literal, Some(constant.span), Some(layout)).unwrap()
+}
+
+pub fn eval_mir_constant<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    constant: &mir::Constant<'tcx>,
+) -> interpret::ConstValue<'tcx> {
```
spernsteiner, line 1528:
```suggestion
// Based on `rustc_codegen_ssa::mir::FunctionCx::eval_mir_constant`
pub fn eval_mir_constant<'tcx>(
    tcx: TyCtxt<'tcx>,
    constant: &mir::Constant<'tcx>,
) -> interpret::ConstValue<'tcx> {
```

src/bin/mir-json-rustc-wrapper.rs
```diff
@@ -49,28 +49,30 @@ impl rustc_driver::Callbacks for GetOutputPathCallbacks {
...
-        let outputs = queries.prepare_outputs().unwrap().peek();
+        let (crate_name, outputs) = {
+            // rustc_session::find_crate_name - get the crate with queries.expansion?
+            let krate = queries.parse().unwrap();
+            let krate = krate.borrow();
+            let crate_name = rustc_session::output::find_crate_name(&sess, &krate.attrs);
+            // get global ctxt then call methods inside of enter() to get
+            let outputs = compiler.build_output_filenames(&sess, &krate.attrs);  //queries.prepare_outputs().unwrap().peek();
+            (crate_name, outputs)
+        };
```
spernsteiner, line 68:
```suggestion
        let (crate_name, outputs) = {
            // rustc_session::find_crate_name - get the crate with queries.expansion?
            let krate = queries.parse().unwrap();
            let krate = krate.borrow();
            let crate_name = rustc_session::output::find_crate_name(&sess, &krate.attrs);
            let outputs = compiler.build_output_filenames(&sess, &krate.attrs);
            (crate_name, outputs)
        };
```
`global_ctxt` is not used here, so I think the "get global ctxt ..." comment is outdated.

spernsteiner (approved):
Looks good overall.  I left a bunch of comments, but most are trivial, like removing old commented-out code.  I tried to leave suggestions for these so you can apply them directly in the github UI.  The non-trivial comments are around `DynKind::DynStar` and test cases for non-literal array sizes (e.g. `const SIZE: usize = 123; ... [i32; SIZE] ...`).

src/analyz/to_json.rs
RyanGlScott: I'm honestly not sure. This came from https://github.com/GaloisInc/mir-json/commit/c13957d2d3017f28a0299d03b385cac88436de4d, so perhaps @m10f could offer some insight here.

src/analyz/ty_json.rs
RyanGlScott: Does this code actually work? When I try it, I get:

```
$ rustc test/conc_eval/consts/unevaluated.rs
error[E0599]: no associated item named `SIZE` found for type parameter `T` in the current scope
  --> test/conc_eval/consts/unevaluated.rs:16:31
   |
16 | fn g<T: HasSize>(xs: [i32; T::SIZE]) -> i32 {
   |                               ^^^^ associated item not found in `T`
   |
   = help: items from traits can only be used if the type parameter is bounded by the trait

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
```

(I might be missing something.)

src/analyz/ty_json.rs
RyanGlScott: Sorry, I think I'm missing something. What `TraintInst` code are you referring to?

src/analyz/ty_json.rs
spernsteiner: Ah, yep, the second one is not (yet) allowed by rustc.  I guess it should look about the same as the first one after mir-json is done monomorphizing, so just testing with `f` should be sufficient.

src/analyz/ty_json.rs
RyanGlScott: Ah, perhaps you are referring to this code?

https://github.com/GaloisInc/mir-json/blob/be4359a4e4361dd2c403e30b51fccc421f8304e4/src/analyz/to_json.rs#L155-L165

I will see what happens if I explicitly rule out `DynKind::DynStar` (hopefully no test cases start failing).

spernsteiner: Yes, that's the code I meant

RyanGlScott: I've split this into separate `Dyn` and `DynStar` cases.

src/analyz/ty_json.rs
RyanGlScott: I've added an `unevaluated.rs` test case in https://github.com/GaloisInc/crucible/pull/1096 that tests `f` above.

src/analyz/to_json.rs
RyanGlScott: I spoke to @m10f. This comment is referring to the fact that this `rust` commit (https://github.com/rust-lang/rust/commit/79db32b64ec946afbef2ed50282f03b71d933480) changed `Term` from an `enum`:

```rs
pub enum Term<'tcx> {
    Ty(Ty<'tcx>),
    Const(Const<'tcx>),
}
```

To a struct:

```rs
pub struct Term<'tcx> {
    ptr: NonZeroUsize,
    marker: PhantomData<(Ty<'tcx>, Const<'tcx>)>,
}
```

And this comment served as a reminder to check if changing this line in `mir-json` from `term: ty::Term::Ty(actual_ty)` to `term: actual_ty.into()` is the proper way to migrate the code. That being said, I took a closer look at the `rust` commit, and I am fairly certain that this is correct. That commit _also_ migrates a similar bit of code [here](https://github.com/rust-lang/rust/commit/79db32b64ec946afbef2ed50282f03b71d933480#diff-ea6f9d55e90333a1f5fc5e202d46862a6a031410831f2df42317ac024f821162R352) in the exact same way:

```diff
-                    term: Term::Ty(output_ty),
+                    term: output_ty.into(),
```

As such, I think the current code is fine. I will remove the `TO REVIEW` comment and land this, but feel free to object if I've overlooked something.

spernsteiner: Yes, that sounds like the correct fix.  The new `Term` is a sort of "packed enum" that uses (I assume) pointer tagging instead of a separate discriminant field to distinguish the `Ty` and `Const` cases, so it's only one word wide instead of two.

PR #62, opened 2023-11-28
Initial CI build

### LINKED ISSUE #42: Add CI
We should add a basic CI setup that ensures that everything builds as expected.

langston-barrett:
Fixes #42

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,77 @@
+name: CI
+
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+
+env:
+  # The CACHE_VERSION environment variable can be updated to force the use of a
+  # new cache if the current cache contents become corrupted/invalid.  This can
+  # sometimes happen when (for example) the OS version is changed but older .so
+  # files are cached, which can have various unintended effects.
+  CACHE_VERSION: 1
+  RUST_TOOLCHAIN: "nightly-2023-01-23"
+
+jobs:
+  lint:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Deps
+      run: |
+        rustup default ${{ env.RUST_TOOLCHAIN }}
+        rustup component add clippy rustc-dev
+    - uses: Swatinem/rust-cache@v2
+      with:
+        cache-on-failure: "true"
+        prefix-key: "${{ env.CACHE_VERSION }}-${{ matrix.os }}"
+    # TODO (#80): Fix lints, use --deny warnings
+    - name: Lint
+      run: cargo clippy --locked
+
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, macos-14]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Deps
+      run: |
+        rustup default ${{ env.RUST_TOOLCHAIN }}
+        rustup component add rustc-dev
+    - uses: Swatinem/rust-cache@v2
+      with:
+        cache-on-failure: "true"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Deps
+      run: |
+        rustup default nightly-2023-01-23
+        rustup component add clippy rustc-dev
+    - uses: Swatinem/rust-cache@v2.7.0
+    # TODO: Fix lints, use --deny warnings
+    - name: Lint
+      run: cargo clippy
+
+  test:
```
RyanGlScott, line 27: Perhaps this should be named "`build`" instead of "`test`"? At present, this job doesn't run any tests.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+    # - run: cargo test --locked --no-run
+    # - run: cargo test 
+    - run: cargo build --bin=mir-json --release
+    - name: Upload binary
+      uses: actions/upload-artifact@v3
+      if: github.repository == 'GaloisInc/${{ env.NAME }}'
+      with:
+        name: "${{ env.NAME }}"
+        path: "target/release/${{ env.NAME }}"
+        if-no-files-found: error
```
RyanGlScott, line 46: Have you confirmed that the resulting binary is actually usable if the toolchain isn't installed?

build.rs
```diff
@@ -6,7 +6,7 @@ fn main() {
     // Add the toolchain lib/ directory to `-L`.  This fixes the linker error "cannot find
     // -lLLVM-13-rust-1.60.0-nightly".
     let out = Command::new("rustup")
-        .args(&["which", "rustc"])
+        .args(["which", "rustc"])
```
RyanGlScott, line 9: This is surprising. What motivated this change?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Deps
+      run: |
+        rustup default nightly-2023-01-23
+        rustup component add clippy rustc-dev
+    - uses: Swatinem/rust-cache@v2.7.0
```
RyanGlScott, line 22: I wonder if we should configure the cache a bit more to make it behave like our typical Haskell GHA setups. After reading the [`rust-cache` README](https://github.com/Swatinem/rust-cache/blob/b1db5f9d5f9ed685de3a2774c3d3b838d2dde462/README.md) a bit, I think we should tweak the following:

* I think we should enable `cache-on-failure` so that build dependencies will still cache even if building `mir-json` itself fails.
* To future-proof this a little bit in case we accidentally corrupt the cache and need to invalidate it, I think it would be worth setting `prefix-key: ${{ env.CACHE_VERSION }}`, where [`CACHE_VERSION`](https://github.com/GaloisInc/cryptol/blob/d602858ea498360126633209d98367654d461995/.github/workflows/ci.yml#L19-L25) is a GHA environment variable.

Similarly for the use of `rust-cache` in the other job below.

RyanGlScott (approved):
Hooray! Thank you so much for doing this.

build.rs
langston-barrett: Clippy! I can revert it if you prefer.

RyanGlScott: Given that there are many more `cargo clippy` warnings that we would need to address anyway, I'm opting to revert this change for now. I've opened https://github.com/GaloisInc/mir-json/issues/80 for the larger issue of resolving `cargo clippy` warnings.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+    - uses: actions/checkout@v4
+    - name: Deps
+      run: |
+        rustup default nightly-2023-01-23
+        rustup component add rustc-dev
+    - uses: Swatinem/rust-cache@v2.7.0
+    # TODO: Fix builds of examples, run tests
+    # - run: cargo test --locked --no-run
+    # - run: cargo test
```
RyanGlScott, line 38: I've opened #81 to track the idea of running the tests in CI.

PR #65, opened 2024-01-30
Update the openssl wrapper crate to get one that understands aarch64 on macOS.

sauclovian-g:
Update the openssl wrapper crate to get one that understands aarch64 on macOS.

(done with: cargo update --package openssl-src)

This is required to build on macos with Apple Silicon.


RyanGlScott (approved):
Thanks!

PR #75, opened 2024-09-09
Explicitly version the JSON schema

### LINKED ISSUE #45: Add a schema version to JSON
On occasion, `mir-json` must make backwards-incompatible changes to the schema it uses for the JSON files that it produces. Ideally, such changes would be readily apparent from the JSON itself so that tools consuming the JSON can fail early, rather than misinterpreting the results of the new schema. To this end, we should add a version number to each JSON file that `mir-json` produces that describes which version of the schema is being used.

RyanGlScott:
We now include a `"version": <N>` entry in the top-level JSON map that gets generated for each MIR JSON file, where `<N>` represents the version of the JSON schema that is used to generate the file. Going forward, any changes to the JSON schema will be accompanied by a corresponding schema version bump. I have also included some logic in `link_crates` to ensure that all of the crates being linked together use the same schema version.

Fixes #45.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,5 +36,20 @@ book](https://doc.rust-lang.org/book/2018-edition/ch01-01-installation.html)).
 
 See the [crux-mir][crux-mir-repo] README for usage instructions.
 
+## JSON schema
+
+`mir-json` and related tools produce MIR JSON files as output, which the
+contain the intermediate MIR code from the compiled program in a
+machine-readable format. Downstream tools are sensitive to the particular
+schema that a MIR JSON file uses, so we explicitly record the version of the
+JSON schema in each MIR JSON file (in the `"version"` field).
+
+Any time that `mir-json` is updated such that the JSON schema must also be
+changed, we will also update the schema version number. The version number is
+represented as a single integer. Any changes to the schema are assumed to be
+backwards-incompatible with previous versions of the schema, so all version
+bumps should be treated as major version bumps. Each change to the schema is
+described in the [`SCHEMA_VERSIONING.md`](SCHEMA_VERSIONING.md) file.
+
 
 [crux-mir-repo]: https://github.com/GaloisInc/crucible/tree/master/crux-mir
--- a/src/link.rs
+++ b/src/link.rs
@@ -5,6 +5,7 @@ use serde_cbor;
 use serde_json;
 
 use crate::lib_util::{self, CrateIndex, InternTable, EntryKind, StringId};
+use crate::schema_ver::SCHEMA_VER;
 
 
 fn read_crates<R: Read + Seek>(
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -36,5 +36,21 @@ book](https://doc.rust-lang.org/book/2018-edition/ch01-01-installation.html)).
...
+
+`mir-json` and related tools produce MIR JSON files as output, which the
+contain the intermediate MIR code from the compiled program in a
+machine-readable format. Downstream tools are sensitive to the particular
+schema that a MIR JSON file uses, so we explicitly record the version of the
+JSON schema in each MIR JSON file (in the `"version"` field).
+
+Any time that `mir-json` is updated such that the JSON schema must also be
+changed, we will also update the schema version number. This version number
+follows [Semantic Versioning](https://semver.org/). As such, an update to the
```
spernsteiner, line 49: I wonder how much benefit we'll actually get out of using semver instead of a single number (with all versions assumed to be incompatible).  Most changes to the schema so far have been non-backwards-compatible.

spernsteiner (commented):
`link.rs` should probably check the versions of its input files and pick a version that's compatible with all of them for the output.  Otherwise you can have this issues: the user builds a package and its dependencies with mir-json v1, then rebuilds the package only with mir-json v2, and `link.rs` ends up mixing v1 and v2 data and wrongly labels it all as v2.

This might require adding the version to the `CrateIndex` part (named `index.cbor` IIRC) of the pre-link representation - the `crate.json` part of that representation is never parsed in full for performance reasons.

README.md
RyanGlScott: Yes, this is a fair point. I don't have any particular examples in mind where having a minor version would be helpful, so I'll change this from a semver to a simple `u64` instead. We can always generalize this later if requiring a major version bump for each change causes too much churn.

RyanGlScott:
I've pushed a reworked version of this PR that uses only a single number for the schema version, and which explicitly checks that all linked crates have the same schema version number. PTAL.

src/link.rs
```diff
@@ -62,13 +62,41 @@ fn collect_roots(
...
+    }
+}
 
 /// Combine the contents of `ocs`, producing a combined JSON crate data object as the result.
 pub fn link_crates<R, W>(inputs: &mut [R], mut output: W) -> serde_cbor::Result<()>
 where R: Read + Seek, W: Write {
     let (indexes, json_offsets) = read_crates(inputs)?;
     let (it, defs, translate) = assign_global_ids(&indexes);
     let roots = collect_roots(&indexes, &translate);
+    let version = check_equal_versions(&indexes)?;
```
spernsteiner, line 99: It might be worth being stricter here and asserting that `version` matches the current `SCHEMA_VER`, or at least checking that it's new enough to have the same set of `EntryKind`s (and other top-level structure, such as `roots`).  Currently, if we remove an `EntryKind` in schema v2, trying to link together a bunch of v1 inputs using the v2 tools will produce an invalid "v1" output that only has tables for the new `EntryKind`s and omits the removed one.

src/link.rs
```diff
@@ -62,13 +62,41 @@ fn collect_roots(
...
+                        "Mismatched schema version numbers ({:?} and {:?})",
+                        version,
+                        index.version,
+                    )));
+                }
+            }
+            Ok(version)
+        },
+        None =>
+            panic!("check_equal_versions: Empty list of crate indexes"),
```
spernsteiner, line 89: I think it would be legal to just return `SCHEMA_VER` here.  Then linking together zero inputs produces a valid empty file matching the current schema.

spernsteiner (approved):

src/link.rs
RyanGlScott: Good point—it doesn't seem terribly useful to have `mir-json` (which is assumed to work with only one schema version) be able to link files that were produced with an older schema. I've reworked this function to check for this as well.

src/link.rs
RyanGlScott: This `None` case no longer exists in the new version of the code.

sauclovian-g:
It occurred to me just now (in the context of the crucible side) that there's a second factor in here, which is the rustc version. Eventually we'd like not to be tied to the one outdated rustc version, but since we don't control what rustc does we might have trouble encoding different rustc's MIR in the same schema version. So I was thinking we might actually want two major version numbers, one for essentially the MIR revision and one for our own JSON version within that. (Then crucible might support the last three JSON schema versions for each of the currently supported MIR revisions or whatever.)

also there were a couple things I'd add to the docs in here but since it's too late I'll just make another pull request.

RyanGlScott:
> Eventually we'd like not to be tied to the one outdated rustc version, but since we don't control what rustc does we might have trouble encoding different rustc's MIR in the same schema version.

I am skeptical that we will ever be able to build a particular version of `mir-json` with anything more than a single `rustc` version. Given that `mir-json` is implemented using the `rustc` API, it's far too coupled to `rustc`'s internals to be able to support multiple `rustc` versions without using a significant amount of `#if`s (which would be very painful).

That being said, I think it would be fine to bump the schema version whenever we upgrade the version of `rustc` that we build against.

sauclovian-g:
In that case the corresponding situation is that we have branches of mir-json for different rustc versions. It's not inconceivable that we might want to do this and maintain more than one at a time. In that case, you don't want to end up in the situation where e.g. the 5.1 branch is using schema version 169 and the 5.2 branch is on schema version 174 after having gone through 170-173, and then some issue comes along where both need to be bumped.

There's no reason we need to add the extra sequence number now, but then we want to remember to add it as soon as we introduce multiple branches or anything of the sort. Which is fine I guess...

Issue #67, opened 2024-03-13
Require defining `{CRUX,SAW}_RUST_LIBRARY_PATH` by default


RyanGlScott:
Currently, `mir-json` effectively has two modes of operation:

1. Compile Rust code against the modified version of the Rust standard libraries that ship with `crux-mir` ([here](https://github.com/GaloisInc/crucible/tree/af9c12732f4ba75558decde1942c5d4eb12833bb/crux-mir/lib)).
2. Compile Rust code against the versions of the Rust standard libraries that come included with `rustc`.

If the `CRUX_RUST_LIBRARY_PATH` environment variable (or, equivalently, `SAW_RUST_LIBRARY_PATH`) is defined when invoking `mir-json`, then mode (1) is chosen. Otherwise, mode (2) is chosen.

The thing is, mode (2) is very rarely what you want. This is because `mir-json` is almost always used in tandem with `crux-mir` or SAW, and as a result, you really want to emit MIR JSON that is compatible with Crucible. The Rust standard library code that comes included with `rustc` does not meet this requirement for two reasons:

* It does not include the custom tweaks needed to avoid code that Crucible cannot handle.
* It is optimized (unlike the custom standard libraries shipped with `crux-mir`), which means that inlining functions might expose lower-level functions that Crucible either cannot handle or does not have overrides for.

However, it is far too easy for a `mir-json` user to forget to define `{CRUX,SAW}_RUST_LIBRARY_PATH` and accidentally end up in mode (2), which can lead to confusing errors that would not happen under mode (1). (See https://github.com/GaloisInc/crucible/issues/1179 for an example where I fell victim to this confusion myself!)

As a result, I'd like to propose that **`mir-json` should require setting `{CRUX,SAW}_RUST_LIBRARY_PATH` by default**, throwing an error message if they aren't set. It is still possible that a user might want to run `crux-mir` in mode (2) for debugging purposes, but they should be required to pass an additional flag or environment variable that acknowledges the user knows the risks that they are getting themselves into.

PR #83, opened 2025-02-28
Add Docker image

### LINKED ISSUE #71: Provide Docker image
We should offer a Docker image that provides all of the `mir-json` tools for a particular `rustc` version. These are included as part of the [`crux-mir` image](https://github.com/GaloisInc/crucible/pkgs/container/crux-mir), but this is a larger image than what may be needed for certain use cases.

RyanGlScott:
Fixes #71.


RyanGlScott:
I have verified that this image builds locally and runs as you would expect it to. Some things left to do:

* This image does _not_ build our forked versions of the Rust standard libraries (currently housed [here](https://github.com/GaloisInc/crucible/tree/90447535c05834fcb2bd7bb0aa64847b788f69bf/crux-mir/lib)). After some thought, I think this is fine. The plan in https://github.com/GaloisInc/crucible/issues/1252 is to move this code to its own repo, and once we do that, we can provide a separate Docker image that includes the compiled versions of the forked standard libraries.
* We should build this image as part of CI and push it to the GitHub Container Registry (ghcr.io). For now, I think it will suffice to do this on each commit to the `master` branch (where the image is tagged with `nightly`), as well as on each time a `git` tag is pushed (where the image is tagged with the name of the `git` tag).
* We should advertise this image in the `mir-json` `README`.

spernsteiner:
> This image does _not_ build our forked versions of the Rust standard libraries

Is mir-json usable for anything without those libraries?  Even `#[no_std]` code is going to need libcore, and using the standard libcore (e.g. `cargo saw-build -Zbuild-std=core`) would be pretty severely broken (IIRC even some basic slice operations are patched in our fork).

RyanGlScott:
> Is mir-json usable for anything without those libraries?

No, not really. (And especially so after https://github.com/GaloisInc/mir-json/issues/67 is fixed.)

I suppose there is a question of whether we want the `mir-json` repo to house a dev image that provides nothing but `mir-json`, for which the repo that houses the fork of the Rust standard libraries could use to build its own image. Alternatively, we could not offer an image in the `mir-json` repo at all and instead provide an image in the other repo that includes all of `mir-json` + the forked standard libraries. I don't have a particularly strong opinion on how we do it.

Yet another alternative would be to move the forked standard libraries into this repo so that both them and `mir-json` are all under the same repo.

spernsteiner:
I have no real preference, except that I don't think we should advertise an image to users that they can't actually use for anything

RyanGlScott:
I have rebased this on top of the changes from https://github.com/GaloisInc/crucible/issues/1252, and I have updated the Docker image to include the translated standard libraries. Next up: a CI job.

.dockerignore
```diff
@@ -0,0 +1,25 @@
+Dockerfile*
+.git
+.gitignore
+
+# Artifacts of running translate_libs.hs
```
qsctr, line 5:
```suggestion
# Artifacts of running translate_libs.sh
```

qsctr (approved):
I am not a CI expert but this looks ok to me

PR #86, opened 2025-03-08
Require `{CRUX,SAW}_RUST_LIBRARY_PATH` or `MIR_JSON_USE_RUSTC_LIBRARY`

### LINKED ISSUE #67: Require defining `{CRUX,SAW}_RUST_LIBRARY_PATH` by default

qsctr:
Fixes #67.

I wasn't really sure how to name `MIR_JSON_USE_RUSTC_LIBRARY`, feel free to suggest something better.

### DIFF
--- a/doc/rustc.md
+++ b/doc/rustc.md
@@ -100,8 +100,8 @@ For the purposes of debugging `mir-json` itself, the environment variable
 `CRUX_RUST_LIBRARY_PATH` or `SAW_RUST_LIBRARY_PATH`, in which case
 `mir-json-rustc-wrapper` will use the vanilla standard library that comes with
 `rustc` instead of a modified version for the compilation process. Since the
-unmodified version is not supported by Crucible, `cargo-crux-test` and
-`cargo-saw-build` **will fail** when `MIR_JSON_USE_RUSTC_LIBRARY` is set.
+unmodified version is not supported by Crucible, the resulting JSON output
+**will not work** with SAW or Crux when `MIR_JSON_USE_RUSTC_LIBRARY` is set.
 
 ## Other binaries
 
--- a/doc/rustc.md
+++ b/doc/rustc.md
@@ -95,6 +95,14 @@ internal environment variables, which most users will not need to care about:
   will export all top-level functions. Otherwise, it will only export those
   functions with a `#[crux::test]` attribute.
 
+For the purposes of debugging `mir-json` itself, the environment variable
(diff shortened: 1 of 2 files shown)


RyanGlScott:
This is great! I think the only thing missing is to document what the `MIR_JSON_USE_RUSTC_LIBRARY` environment variable does in [this documentation](https://github.com/GaloisInc/mir-json/blob/2816dc418494bf586612488ac9982f666e07737e/doc/rustc.md).

qsctr:
Added! I didn't even know that documentation existed until now, heh. I was about to open a separate PR to add a link to it from the README, but I noticed you've done so already in #85 :)

RyanGlScott (approved):

PR #92, opened 2025-03-31
CI: Fix branch name

RyanGlScott:
This repo uses a branch named `master`, not `main`. This fixes incorrect references to `main` in the CI configuration.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   push:
     branches:
-    - main
+    - master
   pull_request:
 
 env:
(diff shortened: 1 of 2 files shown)


spernsteiner (approved):
Seems reasonable, though I suppose the "on merge to master" parts will be particularly annoying to test if anything goes wrong

RyanGlScott:
Yeah, that part is only there since I'm not quite sure how often we want to build the `nightly` tag for the Docker image. We could imagine building that image periodically on a `cron` job, similarly to how the `nightly` versions of Crux-MIR and SAW are built. (I think it remains to be seen what cadence we'll want.)

PR #93, opened 2025-04-02
Automated version of `translate_libs.sh` (`mir-json-translate-libs`)

### LINKED ISSUE #89: Automate generation of `translate_libs.sh` script
Currently, `crux-mir`'s [`translate_libs.sh`](https://github.com/GaloisInc/crucible/blob/96343bf9eee378c2413b2c7d1bed806a8a57731c/crux-mir/translate_libs.sh) script is written by hand. This makes it tedious to maintain, especially when new `rustc` toolchains alter the dependencies used in the Rust standard libraries. We should include a program that generates the `translate_libs.sh` script in a more declarative way. @m10f started work on this with this program:

<details>

```py
# gen_build_sh.py
import sys
import shlex
import os
import shutil
from typing import Iterator

def crate_loc(path: str):
  (parent, _) = os.path.split(path)
  (parent2, parent_name) = os.path.split(parent)
  if parent_name == "src":
    return parent2

  return parent


def has_arg(args: list[str], n: str) -> bool:
  return n in args

def get_args(args: list[str], n: str) -> Iterator[str]:
  for arg_idx in range(0, len(args)):
    arg = args[arg_idx]
    if arg == n and arg_idx + 1 < len(args):
      yield args[arg_idx + 1]

def get_arg(args: list[str], n: str) -> str:
  for x in get_args(args, n):
    return x

  raise RuntimeError(f"Missing required arg: {n} args: {args}")

def parse(inp):
  for line in inp:
    line = str(line)
    if "--target" not in line:
      continue

    line = line.removeprefix("Running ").strip(" ").strip("`")
    args = shlex.split(line)

    crate_name = get_arg(args, "--crate-name")
    edition = None
    if has_arg(args, "--edition=2021"):
      edition = 2021
    if has_arg(args, "--edition=2018"):
      edition = 2018
    if has_arg(args, "--edition=2015"):
      edition = 2015

    cfgs = get_args(args, "--cfg")
    externs = [ x.split("=")[0] for x in get_args(args, "--extern") ]

    # this is kinda brittle but it's not really tagged
    lib_path = next(a for a in args if a.endswith(".rs"))
    crate_path = crate_loc(lib_path)

    yield {
      "edition": edition,
      "crate_name": crate_name,
      "cfgs": cfgs,
      "externs": externs,
      "crate_path": crate_path,
      "lib_path": lib_path,
    }

def copy(crate, srcpath):
  crate_path = crate["crate_path"]
  shutil.copytree(crate_path, os.path.join(srcpath, crate["crate_name"]))

def build_cmd(crate, libdir, srcpath):
  cfg_args = [f"--cfg {shlex.quote(c)}" for c in crate["cfgs"]]
  edition = crate["edition"]
  externs = crate["externs"]
  crate_name = crate["crate_name"]
  crate_path = crate["crate_path"]
  lib_path = crate["lib_path"]
  lib_rel_path = os.path.relpath(lib_path, crate_path)
  target_path = os.path.join(srcpath, crate_name, lib_rel_path)


  extern_args = [f"--extern {e}={os.path.join(libdir, 'lib' + e + '.rlib') }" for e in externs]
  cmd = [
    "mir-json",
    target_path,
    f"--edition={edition}" if edition is not None else "",
    f"--crate-name {crate_name}",
    f"-L {libdir}",
    f"--out-dir {libdir}",
    f"--crate-type rlib",
  ] + cfg_args + extern_args
  return " ".join(cmd)

crates = parse(sys.stdin)
if sys.argv[1] == "copy":
  for crate in crates:
    copy(crate, "lib")

elif sys.argv[1] == "buildscript":
  print("set -e")
  for crate in crates:
    print(f"echo 'Building {crate['crate_name']}...'")
    print(build_cmd(crate, "rlibs", "lib"))
    print()
```
</details>

Where it is intended to be used with `gen_build_sh.py buildscript < log_from_building_an_empty_project`, where "`log_from_building_an_empty_project`" is the result of running something like `cargo build -Z build-std --verbose`.

The script also includes a `copy` mode that can be used to copy the source code for the Rust standard libraries into a central location. This could be used to automate the process of updating `crux-mir/lib` to a new `rustc` toolchain.

qsctr:
This adds a new binary `mir-json-translate-libs`, which automatically invokes `cargo test -Z build-std` to figure out the `mir-json` commands necessary for translating the modified standard libraries for the current target. It also then executes those `mir-json` commands directly, so ~`translate_libs.sh` is now a thin wrapper around `mir-json-translate-libs`, and can be used in the same way as before~ it replaces `translate_libs.sh`. The only difference is we now require the `rust-src` rustup component to be installed, which is necessary for `build-std`.

`mir-json-translate-libs` also supports a `--generate` option, which will output a shell script containing `mir-json` commands instead of executing them directly, like the original plan of #89.

The changes in the `libs/` directory are to line it up with the format that `cargo` expects for library sources, which `mir-json-translate-libs` follows. Namely, the directory names for each library are the package name instead of the crate name, so some `_`s were converted to `-`s. Also, the `rustc-std-workspace-std` package was added. (`translate_libs.sh` previously worked without it because we weren't really using the `rustc-std-workspace-*` packages like `cargo` was. They are a cargo workaround for packages on crates.io that `build-std` depends on, but which themselves depend on packages in `rust-src`. `rustc-std-workspace-X` basically just re-exports everything from `X`, but we were just using `X` directly instead. Now we do it the same way as `cargo` does to follow it more closely.)

(Also, this is my first time actually writing any nontrivial Rust code, so please let me know if I did anything unusual or non-idiomatic.)

Closes #89.


src/bin/mir-json-translate-libs.rs
```diff
@@ -0,0 +1,692 @@
...
+    program: String,
+    args: Vec<String>,
+    env: Vec<(String, String)>,
+}
+
+impl CmdInvocation {
+    fn as_command(&self) -> Command {
+        let mut cmd = Command::new(&self.program);
+        cmd.args(&self.args)
+            .envs(self.env.iter().map(|(k, v)| (k, v)));
```
RyanGlScott, line 280: For my own Rust learning: can you explain why the purpose of the `.map(|(k, v)| (k, v))` call is? At first glance, that looks like a no-op, but perhaps there is a subtle type change going on there.

src/bin/mir-json-translate-libs.rs
```diff
@@ -0,0 +1,692 @@
...
+        units: unit_graph.units.into_iter().map(convert_unit).collect(),
+        roots: unit_graph.roots,
+    };
+
+    // Add our extra libraries to the unit graph.
+
+    let dep_core = custom_graph.get_unit_as_dep("core".into());
+    let dep_compiler_builtins =
+        custom_graph.get_unit_as_dep("compiler_builtins".into());
+    let dep_std = custom_graph.get_unit_as_dep("std".into());
```
RyanGlScott, line 496: Could you factor out the extra libraries into a top-level constant? That way, it is easy to tell at a glance what they are, and it will be simpler to add more later if need be.

src/bin/mir-json-translate-libs.rs
```diff
@@ -0,0 +1,692 @@
...
+            .iter()
+            .collect(),
+            edition: Edition::E2021,
+            linked_libs: vec![],
+            linked_paths: vec![],
+            cfgs: vec![],
+            env: vec![],
+        },
+        vec![dep_compiler_builtins.clone(), dep_core.clone()],
+    );
```
RyanGlScott, line 516: Adding `crucible` as a dependency is a surprisingly manual process. Is there any way to infer this information from `crucible`'s `Cargo.toml` file? (And similarly for `int512`, `byteorder`, etc.)

src/bin/mir-json-translate-libs.rs
```diff
@@ -0,0 +1,692 @@
...
+        .output()
+        .expect("rustc --print target-libdir should run")
+        .stdout;
+    let rlibs_dir: &Utf8Path = std::str::from_utf8(&rlibs_dir)
+        .expect("rustc target-libdir should be UTF-8")
+        .trim_end()
+        .into();
+
+    // Set up a new cargo project for running `cargo test -Z build-std`.
+    eprintln!("Creating empty cargo package...");
```
RyanGlScott, line 335: It might be worth creating this project in a temporary directory using [`tempdir`](https://docs.rs/tempfile/3.19.1/tempfile/fn.tempdir.html). This would avoid needing to have to delete it beforehand in the event of an error, and it would automatically handle cleaning up the project when the `TempDir` is dropped.

translate_libs.sh
```diff
@@ -1,132 +1,3 @@
...
-echo 'Building int512...'
-translate libs/int512.rs --crate-name int512 --extern "core=${RLIBS}/libcore.rlib" --extern "compiler_builtins=${RLIBS}/libcompiler_builtins.rlib"
-
-echo 'Building bytes...'
-translate libs/bytes.rs --edition=2021 --crate-name bytes --extern "core=${RLIBS}/libcore.rlib" --extern "std=${RLIBS}/libstd.rlib" --extern "compiler_builtins=${RLIBS}/libcompiler_builtins.rlib" --extern "crucible=${RLIBS}/libcrucible.rlib"
-
-echo 'Building byteorder...'
-translate libs/byteorder/src/lib.rs --edition=2021 --crate-name byteorder --cfg 'feature="std"' --extern "core=${RLIBS}/libcore.rlib" --extern "std=${RLIBS}/libstd.rlib" --extern "compiler_builtins=${RLIBS}/libcompiler_builtins.rlib"
-
+cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" && mir-json-translate-libs "$@"
```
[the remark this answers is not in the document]
qsctr: Yeah, I kept it around mostly just so it would be a drop-in replacement and the setup process wouldn't change too much.

Technically the bash script does provide the ability of still doing the right thing when the CWD is not the repository root, by virtue of `BASH_SOURCE`, while `mir-json-translate-libs` just does everything in the CWD (since I'm not aware of a good way of finding our way back to the `mir-json` repository in a binary). So if someone runs `dependencies/mir-json/translate_libs.sh` from their `crucible` checkout or something it would still work. Do you think that behavior is worth keeping? (or do you know a good way to do the same thing in a binary?)

src/bin/mir-json-translate-libs.rs
```diff
@@ -0,0 +1,692 @@
...
+}
+
+/// Module to hide field of newtype [CrateName].
+mod crate_name {
+    use std::fmt;
+
+    use serde::Deserialize;
+
+    /// Newtype wrapper around [String] which enforces that all `-` are replaced
+    /// with `_`. Construct with [From<&str>].
```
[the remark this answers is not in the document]
qsctr: It's more of a Rust language requirement rather than `cargo` requirement. I've added an explanation.

src/bin/mir-json-translate-libs.rs
qsctr: There is indeed a subtle type change. `Vec<(String, String)>::iter` returns an `Iterator<Item = &(String, String)>`, but `Command::envs` expects an `IntoIterator<Item = (K, V)>` where `K, V: AsRef<OsStr>`. So the `map` is converting `&(String, String)` to `(&String, &String)`. But pattern-matching on a `&(String, String)` automatically binds each element to `&String`, so we just put them back together in a pair again.

src/bin/mir-json-translate-libs.rs
```diff
@@ -0,0 +1,692 @@
...
+/// Build a `cargo test -Z build-std` command.
+fn cargo_test_cmd(cwd: &Utf8Path, target_triple: &str) -> Command {
+    let mut cmd = Command::new("cargo");
+    cmd.current_dir(cwd)
+        .arg("test")
+        .arg("--target")
+        .arg(target_triple)
+        .arg("-Z");
+    // For some reason we need to explicitly include panic_abort on
+    // wasm32-unknown-unknown.
```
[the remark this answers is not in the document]
qsctr: Yes, I found that we had to add `panic_abort` because I tried running it for `wasm32-unknown-unknown`. It works after I added it.

In the old `translate_libs.sh` we supported setting the target with the `TARGET` environment variable so I supported it here too. We only claimed to have tested it on `wasm32-unknown-unknown` (and the usual host targets), so I tested that here too. I haven't tested it on any other non-host target.

src/bin/mir-json-translate-libs.rs
qsctr: Do you just mean the strings of the library names? Or do you mean the whole `CustomUnit`?

translate_libs.sh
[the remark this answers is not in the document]
RyanGlScott: > Yeah, I kept it around mostly just so it would be a drop-in replacement and the setup process wouldn't change too much.

We control the entire setup process, so I think we should feel emboldened to change if it would simplify things.

> Technically the bash script does provide the ability of still doing the right thing when the CWD is not the repository root, by virtue of `BASH_SOURCE`, while `mir-json-translate-libs` just does everything in the CWD (since I'm not aware of a good way of finding our way back to the `mir-json` repository in a binary). So if someone runs `dependencies/mir-json/translate_libs.sh` from their `crucible` checkout or something it would still work. Do you think that behavior is worth keeping? (or do you know a good way to do the same thing in a binary?)

What if we generalize `mir-json-translate-libs` to take the path to the `libs` directory as an explicit argument? That way, folks could run `mir-json-translate-libs dependencies/mir-json/libs` from a different directory. (If the path isn't specified, we can assume that it's `$(pwd)/libs`.)

src/bin/mir-json-translate-libs.rs
RyanGlScott: I had the strings representing the library names in mind when I wrote that comment.

src/bin/mir-json-translate-libs.rs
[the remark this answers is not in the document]
RyanGlScott: Ah, nice!

I would love if it the CI could check that `wasm32` continues to be supported going forward. I've opened https://github.com/GaloisInc/mir-json/issues/95 (or rather, I transferred an existing issue from the `crucible` repo) to track this.

src/bin/mir-json-translate-libs.rs
qsctr: Unfortunately only `byteorder` has a `Cargo.toml` out of those 4 libraries. Even if we added it for the rest, I think the only information we can easily get from `Cargo.toml` is `crate_name`, `src_path`, and `edition`. I think everything else would require Cargo to do a build, and since these aren't in the existing unit graph, we would have to generate separate unit graphs for them and merge them. Plus we still need to manually add `crucible` as a dependency of `alloc` anyways. So I think it's more trouble than it's worth.

RyanGlScott: OK, fair enough. In that case, could you leave a comment somewhere describing how you figured out what metadata for use for these libraries? My hope is that if we have to do this again in the future (e.g., if we add another extra dependency for some reason), then we can rely on this comment to help guide us rather than figuring it out from scratch.

translate_libs.sh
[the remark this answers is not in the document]
qsctr: I actually already did that, but just didn't document it:
https://github.com/GaloisInc/mir-json/blob/5b0caf1405e08d80468d05c96a861d73284b4ebf/src/bin/mir-json-translate-libs.rs#L313-L319

I was more thinking about where we're putting `rlibs`. But now that I think about it more, I guess there's no particular reason it needs to be in `mir-json`. So maybe it's reasonable to just dump it wherever the user's running `mir-json-translate-libs` from.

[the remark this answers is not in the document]
RyanGlScott: Perhaps we should put `rlibs` alongside the input `libs` directory? Even better if the user can override this location using an `-o` command-line option.

src/bin/mir-json-translate-libs.rs
qsctr: Good idea!

src/bin/mir-json-translate-libs.rs
qsctr: How I figured it out was just by transcribing the `mir-json` commands in the old `translate_libs.sh` :smile:

I'm afraid I don't know how whoever wrote those figured it out in the first place...

src/bin/mir-json-translate-libs.rs
spernsteiner: Might be clearer to make this explicit: `|&(ref k, ref v)| (k, v)`

src/bin/mir-json-translate-libs.rs
```diff
@@ -0,0 +1,701 @@
...
+}
+
+#[derive(Clone, Deserialize)]
+struct UnitGraphDependency {
+    index: usize,
+    extern_crate_name: CrateName,
+}
+
+/// Module to hide field of newtype [CrateName].
+mod crate_name {
```
spernsteiner, line 64: Enforcing the invariant this way is probably a good idea, but honestly, I normally don't bother.  For something that's used only in a fairly small scope (just this file), I don't think it's too bad to rely on people either reading the docs before they use it or copying how it's used in existing examples.  Though I will still use `pub` on fields/methods of private types just as a hint that "other code in the module should generally access these things, but not those ones".

src/bin/mir-json-translate-libs.rs
```diff
@@ -0,0 +1,701 @@
...
+        }
+    }
+}
+
+/// A library that will be compiled by mir-json.
+struct MirJsonLib {
+    target: CustomTargetLib,
+    /// `(x, y)` is passed as `--extern x=liby.rlib`. Usually `x` is the same
+    /// as `y` but not always.
+    dependencies: Vec<(CrateName, CrateName)>,
```
spernsteiner, line 159: Is the `y` part actually required to be a `CrateName`?  IIRC Cargo typically uses `-` in dependency filenames, like `libfoo-00112233.rlib`

src/bin/mir-json-translate-libs.rs
```diff
@@ -0,0 +1,701 @@
...
+        .stdout;
+    let rlibs_dir: &Utf8Path = std::str::from_utf8(&rlibs_dir)
+        .expect("rustc target-libdir should be UTF-8")
+        .trim_end()
+        .into();
+
+    // Set up a new cargo project for running `cargo test -Z build-std`.
+    eprintln!("Creating empty cargo package...");
+    let empty_project_dir =
+        tempdir().expect("temporary directory should be able to be created");
```
spernsteiner, line 354: IMO there's not much value in `do_thing().expect("thing should have worked")` over `do_thing().unwrap()`.  Calling `unwrap()` already sort of implies that you thought it should never fail.

src/bin/mir-json-translate-libs.rs
spernsteiner: Are the strings used in multiple places?  IMO there's not much value in changing `f("foo")` to `const FOO = "foo"; ...; f(FOO)`.  But if the same string is used in multiple places, then it's good to make sure they don't get out of sync.

spernsteiner (commented):
Since you asked about style/idiomaticity, here are a few nitpicks (feel free to ignore if you disagree)

src/bin/mir-json-translate-libs.rs
```diff
@@ -0,0 +1,701 @@
...
+/// [cargo_metadata::Artifact], and [cargo_metadata::BuildScript]. Each
+/// corresponds to some input to `rustc`.
+#[derive(Clone)]
+struct CustomTargetLib {
+    /// `--crate-name`
+    crate_name: CrateName,
+    /// source file argument
+    src_path: Utf8PathBuf,
+    /// `--edition`
+    edition: Edition,
```
spernsteiner, line 134: Seeing this makes me wonder whether `--unit-graph` is the right thing to use for this tool.  The `--build-plan` output gives the full rustc arguments and environment for each step, whereas here it seems like we're reconstructing a bunch of that ourselves.

qsctr: It seems like `--build-plan` is [considered broken and will possibly be removed in the future](https://github.com/rust-lang/cargo/issues/7614). Also, when used in conjunction with `build-std`, Cargo outputs a warning that "`-Zbuild-std` does not fully support `--build-plan`". So I didn't use it.

src/bin/mir-json-translate-libs.rs
qsctr: They are not used in multiple places.

(well, technically they also form part of the source path for each library, but I think those should probably be kept separate.)

src/bin/mir-json-translate-libs.rs
qsctr: Ok. I'll leave this one this way since I already did it, but will keep that in mind for the future!

src/bin/mir-json-translate-libs.rs
qsctr: Ah, good point

src/bin/mir-json-translate-libs.rs
qsctr: I don't necessarily think it will never fail, but rather that there is a chance that it will fail but it's not worth handling the error/doing anything better than panicking. So if it does fail I want the error message to be somewhat more useful than the generic one from `unwrap()`.

doc/rustc.md
```diff
@@ -162,18 +164,53 @@ in the following ways:
...
+    mir-json-translate-libs [OPTIONS] [libs]
+
+ARGS:
+    <libs>    Directory containing custom Rust standard libraries [default: ./libs]
+
+OPTIONS:
+        --generate             Print a shell script instead of actually running the build
+    -h, --help                 Print help information
+    -o, --out-dir <OUT_DIR>    Directory to place rlibs and rlibs_real in [default: next to libs]
+        --target <TARGET>      Rust target triple to configure the libraries for
```
RyanGlScott, line 191: I presume that the default value of `--target` is your host machine's target triple? If so, it would be worth including a `[default: ...]` part of this flag description, similar to the one that `--out-dir` has.

src/bin/mir-json-translate-libs.rs
```diff
@@ -0,0 +1,735 @@
...
+                .long("out-dir")
+                .value_name("OUT_DIR")
+                .help(
+                    "Directory to place rlibs and rlibs_real in \
+                    [default: next to libs]",
+                ),
+            clap::Arg::new("target")
+                .long("target")
+                .value_name("TARGET")
+                .help("Rust target triple to configure the libraries for"),
```
RyanGlScott, line 337: Similarly, say what the default `--target` value is here.

RyanGlScott (approved):
I'm thrilled with the way this turned out. Thanks for working on this!

RyanGlScott:
Also, can you verify that the Dockerfile still builds after your changes?

qsctr:
Good catch, I forgot to update the Dockerfile with the `rust-src` dependency

Issue #72, opened 2024-07-18
Don't use macros to derive `ToJson` impls


RyanGlScott:
Currently, `mir-json` makes use of [two macros](https://github.com/GaloisInc/mir-json/blob/ac740eee106ccc4cf9dfbd80285201b103a0145e/src/analyz/to_json.rs#L399-L443) to automatically derive `ToJson` implementations for several types based on their variant names. Here is one place where such a macro is used:

https://github.com/GaloisInc/mir-json/blob/ac740eee106ccc4cf9dfbd80285201b103a0145e/src/analyz/ty_json.rs#L41

As a result, the `CastKind` `FloatToFloat` will be rendered as the string `"FloatToFloat"` in JSON, `PtrToPtr` will be rendered as `"PtrToPtr"`, and so on.

This approach, while terse, has the downside that if the variant names change between different versions of the `rustc` API, then the JSON schema used in the `ToJson` impls will silently change. This actually matters in practice. For instannce, the version of `rustc` that `mir-json` currently pins against (`nightly-2023-01-23`) has a `CastKind` named `Pointer`, but in version 1.79.0 of `rustc`, `Pointer` is now named `PointerCoercion` instead, which means that downstream tools (e.g., Crucible) would need to parse the corresponding JSON differently. I wouldn't have been aware of this fact unless I bothered to look it up!

To avoid the possibility of nasty surprises like this, I propose that we avoid the use of macros like `basic_json_enum_impl!` and instead hand-write all of the `ToJson` impls, using a `match` expression to ensure that all possible variants are covered. This is more verbose, but it does mean that if future versions of `rustc` change the variant names, then we will be notified immediately when upgrading `rustc` versions, rather than having to play detective after the fact to figure out what changed.

PR #97, opened 2025-04-07
Remove basic_json_impl! and basic_json_enum_impl! macros

sand7000:
Replaced the basic_json_impl! and basic_json_enum_impl! macros with explicit ToJson impls. Enum variants and types are now handled directly in code rather than through macro expansion. This avoids silent changes to the JSON schema if variant names shift between rustc versions.

This should resolve:

https://github.com/GaloisInc/mir-json/issues/72


RyanGlScott (approved):
Thanks, @sand7000!

PR #99, opened 2025-04-08
add rustc-dev component to rust-toolchain.toml

spernsteiner:
This causes rustup to install the component automatically the first time a Rust tool is run in the mir-json directory.


rust-toolchain.toml
```diff
@@ -1,2 +1,3 @@
 [toolchain]
 channel = "nightly-2025-02-16"
+components = ["rustc-dev"]
```
RyanGlScott, line 3: I think this should include `rust-src` as well, as that is now a requirement to run `mir-json-translate-libs` (see #93).

RyanGlScott (approved):

qsctr (approved):

Issue #48, opened 2023-06-30
Publish to crates.io


stefnotch:
It'd be lovely if this tool were published to crates.io, like the majority of Rust tools.

Then the installation instructions might be simpler for users who already have Rust.

RyanGlScott:
I agree. One reason that we haven't yet uploaded `mir-json` to `crates.io` is that any given `mir-json` release would only be buildable with a particular `rustc` nightly, and I could imagine lots of bug reports from confused users who try to install `mir-json` with an unsupported `rustc` version. That being said, we could include a [`rust-toolchain.toml`](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) file to ensure that when a user runs `cargo install mir-json`, `cargo` will download the correct nightly before attempting to build the tool.

RyanGlScott:
Actually, I'm not sure if `rust-toolchain.toml` takes effect with `cargo install` or not. It _would_ take effect if you cloned the `mir-json` repo and built it from source, which would simplify the current story slightly (in that you wouldn't need to manually change your Rust toolchain before building).

stefnotch:
Ah, I see. I suppose a stable mir format would be needed for it to work with multiple rustc versions.

I personally would love anything that simplifies the installation instructions. Even getting rid of a single step is frequently worth it, in my opinion.

RyanGlScott:
Ah, it turns out that we already have a `rust-toolchain.toml` file:

https://github.com/GaloisInc/mir-json/blob/3b538589852282a752e517b69fc61b627cd6d4bd/rust-toolchain.toml#L1-L2

I think that we need to add `components = ["rustc-dev"]` as well, but most of the ingredients are there.

Another thought that popped into my mind: there are several commands inside of `mir-json` that spawn `rustc` subprocesses. If a user installs `mir-json` with one `rustc` version (the intended version) but later changes their toolchain to have a different `rustc` version as the default, will `mir-json`'s `rustc` subprocesses still pick the intended version? Surprisingly, some experimentation suggests that the answer is "yes". I'm not sure how that works exactly, but that is a fortunate coincidence.

spernsteiner:
> If a user installs `mir-json` with one `rustc` version (the intended version) but later changes their toolchain to have a different `rustc` version as the default, will `mir-json`'s `rustc` subprocesses still pick the intended version?  Surprisingly, some experimentation suggests that the answer is "yes".

[...]

PR #100, opened 2025-04-09
Simpler install

sand7000:
This addresses:

https://github.com/GaloisInc/mir-json/issues/48

It adds a script to simplify installation a bit and it adds a RUSTUP_TOOLCHAIN env var to wrapper.rs to make sure the installed tools always use the right toolchain.


install-mir-json.sh
```diff
@@ -0,0 +1,24 @@
...
+
+TOOLCHAIN="nightly-2023-01-23"
+
+# Check for rustup
+if ! command -v rustup >/dev/null 2>&1; then
+  echo "rustup is not installed."
+  echo "Please install rustup by running:"
+  echo "  curl https://sh.rustup.rs -sSf | sh"
+  echo "and update your path:"
+  echo "  source $HOME/.cargo/env"
```
RyanGlScott, line 13: Instead of putting the exact shell commands to run here, I'd recommend pointing folks to the instructions at https://rustup.rs instead, as it actually recommends running a different command than what is listed here. What's more, this would future-proof the script in the event that the https://rustup.rs instructions change later.

README.md
```diff
@@ -17,17 +17,13 @@
...
-2. Next, install a version of `rustc` that works with mir-json.
+2. You will need basic build tools like `cc` to compile mir-json. On Ubuntu this is sufficient:
 
-       $ rustup toolchain install nightly-2023-01-23 --force
-       $ rustup component add --toolchain nightly-2023-01-23 rustc-dev rust-src
-       $ rustup default nightly-2023-01-23
+       $ sudo apt install build-essential
 
-   <!-- Note: when changing to a new nightly, also update `wrapper.rs` -->
+3. Run the mir-json installation script
```
RyanGlScott, line 24: For consistency with other parts of the instructions:

```suggestion
3. Run the mir-json installation script:
```

RyanGlScott (commented):
Thanks for the PR!

I am quite happy with the `RUSTUP_TOOLCHAIN` changes in `wrapper.rs`. I have more mixed feelings about the introduction of a wrapper script, since it's yet another way way to install `mir-json` that we would need to maintain going forward (e.g., remembering to update the toolchain version). That being said, it does replace multiple `cargo`/`rustup` commands with a single script, so I can see the appeal behind it. I won't push back if others are on board.

sand7000:
> I have more mixed feelings about the introduction of a wrapper script, since it's yet another way way to install `mir-json` that we would need to maintain going forward

This is fair although I find maintaining a README to be more error prone and more likely to fall out of date than a testable script. Case in point the Rust book reference in our README was already out of date.

qsctr:
I don't have strong feelings about this. I think the toolchain and component installations can actually be combined into one command, like we document in `wrapper.rs`:
https://github.com/GaloisInc/mir-json/blob/92d2e94908983e8744fb1ac7fa9e9532ef504fa0/src/bin/wrapper.rs#L48
(speaking of which, that line probably should be updated to just `./install-mir-json.sh`, if we decide we are using the script.)

At that point, the shell script really is just turning two commands into one command, so I'm not sure that it's necessary. Also, I think there is some value in having the command for setting up the toolchain and the command for building & installing `mir-json` itself be separate, in case one of them goes wrong. But that mainly comes from past experiences trying to install software with complex installation scripts that fail somewhere in the middle, and not being able to figure out what went wrong, but as you say, this script is simple enough to inspect, so I guess it's fine.

PR #103, opened 2025-04-16
cargo saw-build and saw rustc now output the same size json

sand7000:
After instrumenting the code to count roots, track dependency traversals, and inspect how everything was linked together, it turned out the difference was surprisingly simple.

I ran:

```
RUSTC_WRAPPER=echo cargo saw-build
```
and

```
RUSTC_WRAPPER=echo saw-rustc
```

and noticed that `cargo` was passing the `--edition` flag, while `saw-rustc` was not. That made all the difference.

Once I added `--edition 2021`, the behavior and output matched `cargo`

### DIFF
--- a/src/bin/saw-rustc.rs
+++ b/src/bin/saw-rustc.rs
@@ -25,6 +25,13 @@ fn main() {
         args.push(host_triple().into());
     }
 
+    // This makes saw rustc behavior match the behavior of cargo
+    // without it the json output is much larger than that of cargo
+    if args.iter().all(|a| !a.starts_with("--edition")) {
+        args.push("--edition".into());
+        args.push("2021".into());
+    }
+
     let my_path = PathBuf::from(env::args_os().nth(0).unwrap());
     let wrapper_path = if let Some(dir) = my_path.parent() {
         dir.join("mir-json-rustc-wrapper")


spernsteiner:
I don't think it's a good idea for `saw-rustc` to have a different default edition from regular `rustc` - that's likely to be confusing for users

sand7000:
> I don't think it's a good idea for `saw-rustc` to have a different default edition from regular `rustc` - that's likely to be confusing for users

How would we like to proceed then? Just document the difference between `cargo saw-build` and `saw-rustc`?

RyanGlScott:
Good find, @sand7000!

I tend to agree with @spernsteiner that as a general rule, we should try to keep `saw-rustc` using the same defaults as `rustc` itself whenever possible to adhere to the principle of least surprise. The fact that different language editions produce such wildly different MIR JSON sizes is a bit counterintuitive, but perhaps there's not much we can do about that. I would be in support of adding a blurb about the `--edition` flag to the `mir-json` documentation [here](https://github.com/GaloisInc/mir-json/blob/f7ad1a48e3207264d6948e7bafd6fe82ff4444a9/doc/rustc.md#other-binaries) (next to `crux-rustc`/`saw-rustc`), since I think this is a footgun that many users would be likely to run into.

Issue #94, opened 2023-11-13
Determine why `saw-rustc` produces larger MIR JSON files than `cargo saw-build`


RyanGlScott:
Consider the following program:

```rs
pub fn f() -> u32 {
    42
}

pub fn g() {
    panic!("Don't run me")
}
```

It is conceivable that we would want to verify `f` in SAW but not `g`. Despite this, if you load this program's MIR JSON into SAW using `mir_load_module`, it will crash before you even have a chance to verify any functions:

```
[12:54:23.144] Loading file "/home/ryanscott/Documents/Hacking/Haskell/saw-script/test.saw"
[12:54:23.180] standalone use of `dyn` is not supported: TyDynamic core/73237d41::any[0]::Any[0]::_trait646b6cc91b1abac7[0]
CallStack (from HasCallStack):
  error, called at src/Mir/TransTy.hs:228:25 in crucible-mir-0.1-inplace:Mir.TransTy
  tyToRepr, called at src/Mir/Trans.hs:868:12 in crucible-mir-0.1-inplace:Mir.Trans
  evalCast', called at src/Mir/Trans.hs:936:5 in crucible-mir-0.1-inplace:Mir.Trans
  evalCast, called at src/Mir/Trans.hs:993:30 in crucible-mir-0.1-inplace:Mir.Trans
  evalRval, called at src/Mir/Trans.hs:1219:11 in crucible-mir-0.1-inplace:Mir.Trans
  transStatement, called at src/Mir/Trans.hs:1685:68 in crucible-mir-0.1-inplace:Mir.Trans
  translateBlockBody, called at src/Mir/Trans.hs:1694:28 in crucible-mir-0.1-inplace:Mir.Trans
  registerBlock, called at src/Mir/Trans.hs:1734:10 in crucible-mir-0.1-inplace:Mir.Trans
  genFn, called at src/Mir/Trans.hs:1780:24 in crucible-mir-0.1-inplace:Mir.Trans
  transDefine, called at src/Mir/Trans.hs:2237:30 in crucible-mir-0.1-inplace:Mir.Trans
  transCollection, called at src/Mir/ParseTranslate.hs:76:26 in crucible-mir-0.1-inplace:Mir.ParseTranslate
  translateMIR, called at src/SAWScript/Crucible/MIR/Builtins.hs:358:11 in saw-script-1.0.0.99-inplace:SAWScript.Crucible.MIR.Builtins
```

[...]

PR #98, opened 2025-04-08
Update Rust toolchain to support Rust 1.86 (`nightly-2025-02-16`)

qsctr:
Updates the Rust toolchain from `nightly-2023-01-23` (Rust 1.69) to `nightly-2025-02-16` (Rust 1.86). This means we now support the Rust 2024 language edition.

Since the MIR format is specific to each `rustc` version, the MIR JSON produced after these changes is backwards-incompatible, and the `mir-json` schema version is correspondingly updated to `2`.

There are two main parts to this PR:
1. In `src/`: Updating the version of the `rustc` library that `mir-json` depends on to `nightly-2025-02-16`, and making `mir-json` correctly produce JSON for the MIR defined by that compiler version.
2. In `libs/`: Updating our custom versions of the standard libraries to the vanilla versions that come with Rust 1.86, and then re-patching them as needed so that they can be processed by `mir-json` and `crucible-mir`. Some of the patches are re-applied from the 2023 version, and some are new due to changes in the standard libraries which introduced code that Crucible can't handle. (On the other hand, some of the 2023 patches are no longer necessary, due to either standard library changes or Crucible improvements since 2023.)

The vast majority of the lines changed in this PR are from upstream changes to the standard library sources, which are uninteresting (from the perspective of this PR). So to review the `src/` changes, filter the diff by that path, and to review the `libs/` changes, look at the individual patch commits.

The `crucible` and `saw-script` changes which make use of this new `mir-json` version are in GaloisInc/crucible#1396 and GaloisInc/saw-script#2323 respectively.

The authors of this PR are @spernsteiner, @RyanGlScott, and myself.

Closes #78.

PR #105, opened 2025-05-06
`doc/rustc.md`: Suggest using `--edition 2021` or later with `{crux,saw}-rustc`

RyanGlScott:
This adds a suggestion in the documentation to use `--edition 2021` or later, as `--edition 2015` has been observed to produce undesirably large MIR JSON files.

Fixes https://github.com/GaloisInc/mir-json/issues/94. Supersedes https://github.com/GaloisInc/mir-json/pull/103.

### DIFF
--- a/doc/rustc.md
+++ b/doc/rustc.md
@@ -106,13 +106,24 @@ unmodified version is not supported by Crucible, the resulting JSON output
 ## Other binaries
 
 Besides the main binaries above, `mir-json` also provides a variety of other
-binaries for specialized purposes:
+binaries for specialized purposes.
+
+### `cargo` subcommands
 
 * `cargo-mir-json`: This invokes `cargo rustc`, but replacing `rustc` with
   `mir-json`.
+
+### `rustc` replacements
+
 * `crux-rustc`: A helper that invokes `mir-json-rustc-wrapper` the same way that
   `cargo-crux-test` would run it. This is useful for testing a single file,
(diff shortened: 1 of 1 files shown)


spernsteiner (approved):
These edits seem fine, but did we ever figure out the actual root cause of #94?  Edition changes are usually pretty surface-level, so I wouldn't expect `--edition 2021` to make such a big difference if everything else is working properly.

RyanGlScott:
I'm afraid I have no idea why changing the edition matters.

For what it's worth, the `1.86` toolchain we are upgrading to in #98 sets the default edition to `2021`, so this will become much less of a concern going forward. If you'd prefer, we could also just leave out these edits.

spernsteiner:
> the `1.86` toolchain we are upgrading to in #98 sets the default edition to `2021`

The default hasn't changed in my local 1.86 toolchain:
```sh
$ rustc --version
rustc 1.86.0 (05f9846f8 2025-03-31)
$ rustc --help
...
        --edition 2015|2018|2021|2024
                        Specify which edition of the compiler to use when
                        compiling code. The default is 2015 and the latest
                        stable edition is 2024.
...
```
Changing the default edition in `rustc` would be a breaking change.  The idea from the original editions RFC was that `cargo new` should default to the latest edition, but all other tools should keep defaulting to 2015 for backward compatibility.

RyanGlScott:
Oh wow, I completely misread the `--help` output for `--edition`—my mistake!

PR #108, opened 2025-05-09
Record lang items and their original IDs in the JSON output

RyanGlScott:
`mir-json` now renders `DefId`s for [lang items](https://doc.rust-lang.org/nightly/unstable-book/language-features/lang-items.html) as `$lang/0::<lang_item_name>` so that we can refer to lang items directly by name. This is useful for wiring in definitions directly into `crucible-mir`, but it has the downside that SAW users will have a more difficult time looking up lang item names in SAW. For instance, one could not simply look up `core::option::Option` anymore; instead, one has to look up `$lang::Option`, which is less intuitive.

To make the lives of SAW users a little easier, this adds an additional `lang_items` field to the MIR JSON output that records each lang item `DefId` along with its original, fully qualified `DefId`. For now, we only record the `DefId`s for ADT definitions, as these are the only names for which we bother to replace their `DefId`s with `$lang::<lang_item_name>` at all. Moreover, we only do this for ADT _definitions_, not ADT _instances_, as SAW users are almost certainly going to user the former instead of the latter when looking up names.

-----

Here are the corresponding `crucible` and `saw-script` branches which make use of these changes:

* `crucible`: https://github.com/GaloisInc/crucible/tree/crux-mir-rust-1.86-lang-item-map
* `saw-script`: https://github.com/GaloisInc/saw-script/tree/saw-rust-1.86-lang-item-map

Once this PR is merged, I will merge the relevant `crucible` and `saw-script` changes as well.

### DIFF
--- a/src/analyz/ty_json.rs
+++ b/src/analyz/ty_json.rs
@@ -83,37 +83,46 @@ fn rename_as_lang_item(tcx: TyCtxt, def_id: hir::def_id::DefId) -> Option<LangIt
     }
 }
 
-/// Given a `DefId`, return a pair containing:
+/// Render the given `DefId` into a `String` of the form
+/// `<crate_name>/<disambig>::<defpath>`. The "orig" in the name
+/// `orig_def_id_str` indicates that this will always pick the _original_
+/// `DefId` for rendering purposes, even for
+/// [lang items](https://doc.rust-lang.org/unstable-book/language-features/lang-items.html).
+/// See [`lang_item_def_id_str`] for a variant of this function that renders
+/// lang items differently.
 ///
-/// * A `String` containing the rendered version of the `DefId` in the form
-///   `<crate_name>/<disambig>::<defpath>`.
-///
-/// * An `Option<String>`. If the `DefId` corresponds to a
-///   [lang item](https://doc.rust-lang.org/unstable-book/language-features/lang-items.html),
-///   then this with be a `Some(id_str)` value, where `id_str` will be of the
-///   form `$lang/0::<lang_item_name>`. Otherwise, this will be `None`.
-///
-/// Most of the time, you will want to use [`def_id_str`] instead, which
-/// computes a single `String`. This function is useful for when you need to
-/// use both `String`s.
-pub fn def_id_strs(
+/// Usually, you will want to use [`def_id_str`] instead, which renders the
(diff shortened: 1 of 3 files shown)


qsctr (approved):

src/analyz/ty_json.rs
```diff
@@ -83,21 +83,51 @@ fn rename_as_lang_item(tcx: TyCtxt, def_id: hir::def_id::DefId) -> Option<LangIt
...
+/// [lang item](https://doc.rust-lang.org/unstable-book/language-features/lang-items.html),
+/// then this function will return the lang item's rendered `String`. Otherwise,
+/// this will fall back to the original, fully qualified path when rendering the
+/// `DefId`.
+pub fn def_id_str(
+    tcx: TyCtxt,
+    def_id: hir::def_id::DefId,
+) -> String {
+    let (orig_def_id_str, lang_item_def_id_str) = def_id_strs(tcx, def_id);
+    lang_item_def_id_str.unwrap_or(orig_def_id_str)
```
spernsteiner, line 130: Seems a little unfortunate to allocate and build the `orig_def_id_str` even when we don't need it, but I guess it's not that big of a deal

spernsteiner (approved):
Looks good.  I'm a bit surprised there's no need to mention the lang items table in `link.rs`, but I guess it must work, since you've got that Crucible branch based on this.

src/analyz/ty_json.rs
RyanGlScott: I could split up `def_id_strs` into two functions, one that computes the original `String` and another that computes the lang item `Option<String>`. Then, we could compute the `Option<String>` first, call `unwrap_or_else` on it, and then compute the original `String` if the `Option` is a `None`. That way, we wouldn't need to allocate/build the original `String` unless it's absolutely necessary.

RyanGlScott:
> I'm a bit surprised there's no need to mention the lang items table in `link.rs`

I was surprised too, but it turns out that `link.rs` is written in a way that it is insensitive to the contents of the table entries themselves (with the exception of `roots`).

src/analyz/ty_json.rs
RyanGlScott: I've pushed this idea, which should hopefully reduce allocations in the common case of non-lang items.

Issue #107, opened 2025-05-06
Come up with a better story for `-Zub-checks`


RyanGlScott:
By default, recent versions of Rust insert [undefined behavior checks](https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/ub-checks.html) into its MIR code that assert various properties at runtime (e.g., if a null pointer dereference occurs where it shouldn't). Unfortunately, these checks tend to be implemented in terms of pointer-to-integer casts, which is something that `crucible-mir` does not currently support. What's more, these assertions are generated inline into the code rather than being guarded behind functions, so it is not straightforward to override these checks.

`crux-mir` works around the issue by [invoking `mir-json` with `-Zub-checks=false`](https://github.com/GaloisInc/crucible/blob/e703d3014c50a999d3913460dcd99d17ab4f1e9f/crux-mir/src/Mir/Generate.hs#L75) to disable these checks. We might consider always enabling that flag, although that might be a step too far. Alternatively, we could include a disclaimer in the documentation that you may need to disable `ub-checks` when compiling Rust code to avoid the pitfalls described here.

qsctr:
Is it "a step too far" because there might be use cases of `mir-json` where we want the UB checks?

RyanGlScott:
Generally, we prefer not to override the default `rustc` settings whenever possible to avoid violating the principle of least surprise. That being said, it seems doubtful that we will be able to support any of the UB checks for the time being, and this will be a very common problem for our users (i.e., the source of many future bug reports) going forward. As a result, I'm hesitant to definitely claim this is a step too far—overriding this particular default setting may very well be justified here.

sauclovian-g:
Are the checks stylized enough to pattern-match on reliably? It would be nice to capture them and turn them into assertions/specification material if it can be done. Though that'll obviously take some work...

RyanGlScott:
Probably not, sadly. [Here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=22d6c4bc09a52f7bcdc27750d6e7c915) is an example from a relatively small program, which compiles to the following MIR:

<details>

```rs
fn f3() -> u32 {
    let mut _0: u32;
    let mut _1: u32;
    let mut _2: u32;
    let mut _3: *mut u32;
    let mut _4: *mut u32;
    let mut _5: *mut u32;
    let mut _6: *const ();
    let mut _7: usize;
    let mut _8: usize;
    let mut _9: usize;
[...]

PR #109, opened 2025-05-09
Invoke `rustc` with `-Z ub-checks=false`

RyanGlScott:
See the surrounding comments (and the related discussion in https://github.com/GaloisInc/mir-json/issues/107) for why we do this.


qsctr (approved):

spernsteiner (approved):
Looks good.  I think it makes sense to omit this adjustment from the `mir-json` binary, which is more of an internal development/testing tool.

PR #114, opened 2025-05-19
`mir-json-translate-libs`: Create `out_dir` before creating temporary cargo package if `--keep-temp-build` is on, otherwise create temporary package in `env::temp_dir()`

qsctr:
Creating it in `out_dir` would fail if `out_dir` doesn't exist yet. And we don't want to `mkdir -p outdir` because we shouldn't create any files permanently if the user is doing `--generate`. Putting it in `cwd` should be fine because it is temporary anyways.

### DIFF
--- a/src/bin/mir-json-translate-libs.rs
+++ b/src/bin/mir-json-translate-libs.rs
@@ -16,7 +16,6 @@ use std::{
     env, fmt, fs,
     io::BufReader,
     os::unix,
-    path::Path,
     process::{Command, Stdio},
 };
 
@@ -434,7 +433,7 @@ fn main() {
                 .long("keep-temp-build")
                 .help(
                     "Persist the temporary cargo package created to run \
-                    `cargo test -Z build-std`",
+                    `cargo test -Z build-std` to out-dir",
                 ),
             clap::Arg::new("copy-sources")
                 .long("copy-sources")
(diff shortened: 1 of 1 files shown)


qsctr:
It used to be there, but then when I added the `--keep-temp-build` option to keep it around after the program ends (for debugging purposes), I wanted it to be in a more convenient location.

RyanGlScott:
OK, but surely that is a separate concern? I propose that if `--keep-temp-build` is enabled, then avoid deleting the temporary directory, but also log the location of the temporary directory so that users can find it later.

qsctr:
We do log the location, but I moved it because I got tired of copy-pasting and cd-ing :)

How about if it creates it in `tempdir` when `--keep-temp-build` is off, and in `out_dir` when `--keep-temp-build` is on? (In that case, I think it's fine to `mkdir -p out_dir`.)

RyanGlScott:
> We do log the location, but I moved it because I got tired of copy-pasting and cd-ing

OK. If it's that burdensome, then I won't push back on putting it in a different directory. I only suggest `temp_dir` because that is a standard location for these sorts of tasks.

Having said that, feel free to put the directory wherever it makes sense. As long as it's easy for users to find, I'm happy.

RyanGlScott (approved):

qsctr:
I ended up doing what I described in the previous comment, since I think that's a better approach.

RyanGlScott (approved):

Issue #115, opened 2025-05-20
Consistent `.linked-mir.json` path


qsctr:
It would be nice if `cargo saw-build` had the ability to always put the `.linked-mir.json` file in the same place in a project, independent of target or cargo hash.

This would be helpful for SAW proofs. Currently, when the Rust code is changed in certain ways like adding a dependency, the crate hash becomes different, and the SAW proof needs to be updated to refer to the new `.linked-mir.json`. A bigger issue is that, since the target is part of the `.linked-mir.json` path, if people on different operating systems are working on the same SAW proof, they cannot have a `mir_load_module` in the SAW script that works for everyone, unless either someone cross-compiles, or everyone agrees to copy the `.linked-mir.json` path to a platform-independent location after running `cargo saw-build`. So this copying to platform-independent location should probably just be done by `cargo saw-build` directly.

This is related to #59 (@spernsteiner mentioned the same idea in https://github.com/GaloisInc/mir-json/issues/59#issuecomment-1808857371).

RyanGlScott:
See also #59, which aims to solve a similar problem.

sauclovian-g:
This is what I put in the Formal VerSo build:
```
saw-build:
        cargo saw-build --target wasm32-unknown-unknown --release
        cp target/*/release/deps/*.linked-mir.json target/linked-mir.json
```
which works well enough (if the glob turns up multiple matches, `cp` will fail)

PR #117, opened 2025-05-22
`mir-json-translate-libs`: Make it easier to generate MIR for cross-compilation targets

qsctr:
Rust lets you easily run `rustc` for a non-host target by doing `rustup target add` to your toolchain. But for many targets, compilation won't actually work all the way, because `rustc` expects you to have a linker for the target platform installed, which `rustup` does not provide. However, if you are just trying to generate MIR for a certain target, this shouldn't be a requirement.

In `mir-json-translate-libs`, we run `cargo test -Zbuild-std` on a new cargo project with the appropriate target to determine how to build the standard libraries. If we don't have a linker for the given target, it will fail at the end when it tries to link the test binary, even though at that point we will have already collected all the information we need about how to compile the standard libraries. So we tell `rustc` to use a dummy linker (in this case `true`) to "link" the test binary, and since we pass `--no-run`, `cargo test` will not try to run the nonexistent test binary and will immediately succeed after invoking `true`. This way, the user does not need to have an actual cross-compilation toolchain installed just to set up and run `mir-json` with a custom target.

### DIFF
--- a/src/bin/mir-json-translate-libs.rs
+++ b/src/bin/mir-json-translate-libs.rs
@@ -518,7 +518,17 @@ fn main() {
     eprintln!("Running cargo test...");
     let mut cargo_test_child = {
         let mut cmd = cargo_test_cmd(&empty_project_dir, target_triple);
-        cmd.arg("--message-format").arg("json").arg("--no-run");
+        cmd.arg("--message-format").arg("json")
+            .arg("--no-run")
+            // Hack: When cross-compiling, cargo test will fail without an
+            // appropriate linker installed. But we only care about compiling
+            // the std libraries to rlibs, which rustc can do on its own. So we
+            // pass a dummy linker to get cargo test to succeed at the end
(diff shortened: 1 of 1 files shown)


RyanGlScott (approved):
This seems plausible. Out of curiosity, which cross-compilation target requires this? I don't recall needing this for `wasm32-unknown-unknown`, for instance, but perhaps that's different enough that it wouldn't require this.

spernsteiner (approved):

qsctr:
I was trying to compile for `x86_64-unknown-linux-gnu` on an `aarch64-apple-darwin` host. (I had received a piece of SAW script that had `mir_load_module "target/x86_64-unknown-linux-gnu/..."` in it, and I wondered if I could work around #115 by doing `cargo saw-build --target x86_64-unknown-linux-gnu` on the accompanying Rust code. Alas, in the end I still ended up with a different hash than the `.linked-mir.json` in the SAW script.)

PR #120, opened 2025-05-23
libs: Patch `[T; N]::try_from(Vec<T, A>)`

qsctr:
This solution is suggested by @spernsteiner

(he described it to me verbally so any errors are mine)

### DIFF
--- a/libs/alloc/src/vec/mod.rs
+++ b/libs/alloc/src/vec/mod.rs
@@ -4063,7 +4063,12 @@ impl<T, A: Allocator, const N: usize> TryFrom<Vec<T, A>> for [T; N] {
         // We checked earlier that we have sufficient items.
         // The items will not double-drop as the `set_len`
         // tells the `Vec` not to also drop them.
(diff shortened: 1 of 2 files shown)


libs/alloc/src/vec/mod.rs
```diff
@@ -4063,7 +4063,12 @@ impl<T, A: Allocator, const N: usize> TryFrom<Vec<T, A>> for [T; N] {
         // We checked earlier that we have sufficient items.
         // The items will not double-drop as the `set_len`
         // tells the `Vec` not to also drop them.
-        let array = unsafe { ptr::read(vec.as_ptr() as *const [T; N]) };
-        Ok(array)
+        // let array = unsafe { ptr::read(vec.as_ptr() as *const [T; N]) };
+        // Ok(array)
```
RyanGlScott, line 4067: I'd favor either removing this code (as it's no longer used) or prefacing this with a comment saying why it's commented out.

RyanGlScott (approved):

PR #122, opened 2025-06-17
`vtable_descriptor_for_cast`: Support vtable casts involving structs

RyanGlScott:
Previously, `vtable_descriptor_for_cast` only supported casts where the from/to types were references or raw pointers. Besides these two types, it is also possible to have unsized coercions between structs where exactly one of the struct's underlying fields differ and require a reference/pointer cast. This is the sort of cast seen in the example in GaloisInc/crucible#1433.

This adds support for struct-related vtable casts by extending `vtable_descriptor_for_cast` as needed. When the from/to types are structs, we recursively walk over the fields of the structs to find a pair of mismatched field types, which we use to compute the vtable to use in the cast.

Fixes one half of GaloisInc/crucible#1433. (The other half will be on the `crucible-mir` side.)

### DIFF
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -95,32 +95,6 @@ fn vtable_descriptor_for_cast<'tcx>(
         return None;
     }
 
-    fn vtable_descriptor_for_ref_or_ptr_cast<'tcx>(
-        tcx: TyCtxt<'tcx>,
-        old_pointee: ty::Ty<'tcx>,
-        new_pointee: ty::Ty<'tcx>,
-    ) -> Option<ty::PolyTraitRef<'tcx>> {
-        if !tcx.is_sized_raw(ty::TypingEnv::fully_monomorphized().as_query_input(old_pointee)) {
-            // We produce a vtable only for sized -> TyDynamic casts.
-            return None;
-        }
-
-        // Relevant code: rustc_codegen_ssa::meth::get_vtable
-        let trait_ref = match *new_pointee.kind() {
-            ty::TyKind::Dynamic(ref preds, _, _) =>
-                preds.principal().map(|pred| pred.with_self_ty(tcx, old_pointee)),
-            _ => return None,
-        };
-        let trait_ref: ty::PolyTraitRef = match trait_ref {
-            Some(x) => x,
-            // If there's no trait ref, it means the `Dynamic` predicates contain only auto traits.
-            // The vtable for this trait object is empty.
-            // TODO: handle this better (currently the output omits the "vtable" field)
-            None => return None,
-        };
-        Some(trait_ref)
-    }
-
     fn instantiate_field_ty<'tcx>(
         tcx: TyCtxt<'tcx>,
         field_def: &ty::FieldDef,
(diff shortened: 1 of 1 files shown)


src/analyz/mod.rs
```diff
@@ -95,37 +95,85 @@ fn vtable_descriptor_for_cast<'tcx>(
...
+        )
+    }
+
+    match (*old_ty.kind(), *new_ty.kind()) {
+        // The two base cases: we are casting between two references or two raw
+        // pointers.
+        (ty::TyKind::Ref(_, old_pointee, _), ty::TyKind::Ref(_, new_pointee, _)) =>
+            vtable_descriptor_for_ref_or_ptr_cast(tcx, old_pointee, new_pointee),
+        (ty::TyKind::RawPtr(old_pointee, _), ty::TyKind::RawPtr(new_pointee, _)) =>
+            vtable_descriptor_for_ref_or_ptr_cast(tcx, old_pointee, new_pointee),
```
spernsteiner, line 141:
```suggestion
        (ty::TyKind::Ref(_, old_pointee, _), ty::TyKind::Ref(_, new_pointee, _)) |
        (ty::TyKind::RawPtr(old_pointee, _), ty::TyKind::RawPtr(new_pointee, _)) =>
            vtable_descriptor_for_ref_or_ptr_cast(tcx, old_pointee, new_pointee),
```
And then you can move the code for `vtable_descriptor_for_ref_or_ptr_cast` into this case of the match

src/analyz/mod.rs
```diff
@@ -95,37 +95,85 @@ fn vtable_descriptor_for_cast<'tcx>(
...
+
+        // It is also possible to cast between two arbitrary struct values,
+        // provided that each struct type differs in exactly one field (or a
+        // field within another struct field) that supports a vtable descriptor
+        // cast. (e.g., casting from Arc<Foo> to Arc<dyn Bar>).
```
spernsteiner, line 146: Might be worth noting the connection to `std::ops::CoerceUnsized`, which is the marker trait that enables this kind of cast

src/analyz/mod.rs
```diff
@@ -95,37 +95,85 @@ fn vtable_descriptor_for_cast<'tcx>(
...
+                let old_field_ty = instantiate_field_ty(tcx, old_field_def, old_args);
+                let new_field_ty = instantiate_field_ty(tcx, new_field_def, new_args);
+                if old_field_ty != new_field_ty {
+                    unequal_fields_count += 1;
+                    vtable = vtable_descriptor_for_cast(mir, kind, old_field_ty, new_field_ty);
+                }
+            }
+
+            // If we have found exactly one pair of unequal fields, then use the
+            // resulting vtable descriptor cast.
```
spernsteiner, line 172: Should note that we expect to find exactly one pair of unequal fields in all practical cases.  [Here's an counterexample](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=a7accacdd6e5f58c834528ddfeceeafa) that has two unequal fields, but (1) this seems to be special-cased to `PhantomData` only, and (2) it would be unusual to use `PhantomData` with `T` when `T` is mentioned elsewhere in the struct (there must be a non-phantom field like `*const T` somewhere so `CoerceUnsized` has somewhere to apply the coercion).

src/analyz/mod.rs
```diff
@@ -95,37 +95,85 @@ fn vtable_descriptor_for_cast<'tcx>(
...
 
-    if !tcx.is_sized_raw(ty::TypingEnv::fully_monomorphized().as_query_input(old_pointee)) {
-        // We produce a vtable only for sized -> TyDynamic casts.
-        return None;
+        // Relevant code: rustc_codegen_ssa::meth::get_vtable
+        let trait_ref = match *new_pointee.kind() {
+            ty::TyKind::Dynamic(ref preds, _, _) =>
+                preds.principal().map(|pred| pred.with_self_ty(tcx, old_pointee)),
+            _ => return None,
+        };
```
spernsteiner, line 113: Eventually, for custom DST support, we'll need to update this to look into the last field of `TyKind::Adt`, so it returns the correct `TraitRef` (our stand-in for a vtable pointer, in this code) when `new_pointee` is e.g. `ArcInner<dyn Trait>`.  I'm not sure if we'll need a similar change elsewhere for array->slice unsizing - my guess is no, since mir-json just emits `Unsize` for those.

src/analyz/mod.rs
samcowger: Even though it's unusual, wouldn't we still want to support the `PhantomData` edge case, as in the first half of your linked example? I ask because `Arc` actually does hold two unequal fields - a `NonNull<ArcInner<T>>` and a `PhantomData<ArcInner<T>>` - and I think that in order to support casts of e.g. `Arc<Concrete>` to `Arc<dyn Trait>`, we need to accommodate this pattern.

spernsteiner: Wow! I didn't think there was any reason to use `PhantomData<T>` when `T` is already mentioned in the type, but apparently it's needed for dropck: https://github.com/rust-lang/rust/blob/f3662275e45271b6f0f1d30d572a21e19f3261a1/src/libcore/ptr.rs#L2480.  So we do need to handle `PhantomData` here after all.

samcowger: I've tried to address this in #132.

PR #125, opened 2025-06-23
Proc macro for building crux-mir-comp MethodSpecs

spernsteiner:
This adds the `crucible_spec_macro` library, defining the `#[crux_spec_for(f)]` proc macro, which can be applied in place of `#[crux::test]` to convert a test into a specification for compositional verification.


spernsteiner:
Marked as draft for now because I still need to set up mir-json-generate-libs to build the new library.  The library itself is ready for review.

spernsteiner:
As this is a proc macro crate, it can't contain anything other than proc macros.  So maybe we should name it `crucible_proc_macros`.

Note that there are already some types and functions specific to crux-mir-comp in the main `crucible` library - they're just unusable without the additional crux-mir-comp overrides.  We could split these out, but I'm not sure it's worth it.

RyanGlScott (approved):

PR #138, opened 2025-07-03
Expose appropriate drop method in traits and vtables

samcowger:
See code comments for details. In short, this allows us to manually obtain and call an appropriate drop function for a trait object in situations where rustc doesn't do so for us.

### DIFF
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -794,7 +794,8 @@ fn emit_trait<'tcx>(
     // glue for trait objects, we always include a drop method in emitted traits
     // (and vtables - see `build_vtable_items` below), and we always put it at
     // the beginning of the method list so that `crucible-mir` can find and call
-    // it manually.
+    // it manually. We adjust `InstanceKind::Virtual` indices to account for
+    // this extra method in `ty_json::adjust_method_index`.
     if let Some(tref) = trait_ref {
         let dyn_ty = tref.self_ty();
         let drop_inst = ty::Instance::resolve_drop_in_place(tcx, dyn_ty);
@@ -1135,7 +1136,8 @@ fn build_vtable_items<'tcx>(
     // glue for trait objects, we always include a drop method in emitted
     // vtables (and traits - see `emit_trait` above), and we always put it at
     // the beginning of the method list so that `crucible-mir` can find and call
-    // it manually.
+    // it manually. We adjust `InstanceKind::Virtual` indices to account for
+    // this extra method in `ty_json::adjust_method_index`.
     let concrete_ty = trait_ref.self_ty();
     let drop_inst = ty::Instance::resolve_drop_in_place(tcx, concrete_ty);
     let drop_def_id = drop_inst.def.def_id();
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -818,8 +818,17 @@ fn emit_trait<'tcx>(
         // no principal trait, only marker traits. For example, `dyn Debug +
         // Send + Sync`'s principal trait would be `Debug`, but `dyn Send +
         // Sync` would have no principal trait, because `Send` and `Sync` are
-        // marker traits.
-        panic!("No trait ref for {:?}?", ti);
+        // marker traits. For now, we've decided not to generally support this
+        // case, because it doesn't seem common or useful.
+        eprintln!("warning: no trait ref for {:?}?", ti);
+        // This assert prevents a potential correctness issue.  On drop,
+        // `crucible-mir` will call the first method in the vtable, expecting it
+        // to be `fn drop_in_place(ptr: *mut dyn Trait)`.  Since we don't emit a
+        // drop function here, a different method could end up in that slot, and
+        // `crucible-mir` would call it if it had a matching Crucible type
+        // signature. We prevent this by asserting that there are no other
+        // methods.
+        assert_eq!(methods.len(), 0);
     }
 
     for &m in methods {
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -824,7 +824,7 @@ fn emit_trait<'tcx>(
 
     for &m in methods {
         // `m` is `None` for methods with `where Self: Sized`.  We omit these from the vtable, and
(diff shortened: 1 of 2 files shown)


src/analyz/mod.rs
```diff
@@ -777,12 +777,50 @@ fn emit_trait<'tcx>(
...
+            "signature": drop_sig.to_json(ms),
+        }));
+    } else {
+        // I believe the only way this can happen is if this trait instance was
+        // derived from predicates (`TraitInst::from_dynamic_predicates`) with
+        // no principal trait, only marker traits. For example, `dyn Debug +
+        // Send + Sync`'s principal trait would be `Debug`, but `dyn Send +
+        // Sync` would have no principal trait, because `Send` and `Sync` are
+        // marker traits.
+        panic!("No trait ref for {:?}?", ti);
```
spernsteiner, line 821:
```suggestion
        // I believe the only way this can happen is if this trait instance was
        // derived from predicates (`TraitInst::from_dynamic_predicates`) with
        // no principal trait, only marker traits. For example, `dyn Debug +
        // Send + Sync`'s principal trait would be `Debug`, but `dyn Send +
        // Sync` would have no principal trait, because `Send` and `Sync` are
        // marker traits.
        eprintln!("warning: no trait ref for {:?}?", ti);
        // This assert prevents a potential correctness issue.  On drop,
        // crucible-mir will call the first method in the vtable, expecting it to be
        // `fn drop_in_place(ptr: *mut dyn Trait)`.  Since we don't emit a drop
        // function here, a different method could end up in that slot, and
        // crucible-mir would call it if it had a matching Crucible type signature.
        // We prevent this by asserting that there are no other methods.
        assert_eq!(methods.len(), 0);
```
I know I suggested putting a `panic!` here, but I've changed my mind.  `Box<dyn Send>` is perfectly valid, even if I don't see any real use for it, and our intent in general is that mir-json should not panic on any valid Rust code.  So I think the right thing to do here for now is just print a warning, and in the future we can implement this case properly if it's needed (which might require some changes to `TraitInst`).

src/analyz/mod.rs
```diff
@@ -1088,7 +1126,26 @@ fn build_vtable_items<'tcx>(
     let trait_ref = tcx.instantiate_bound_regions_with_erased(trait_ref);
     let methods = tcx.vtable_entries(trait_ref);
 
-    let mut parts = Vec::with_capacity(methods.len());
+    // I'm not sure of the circumstances in which `methods` would or would not
+    // contain a `ty::vtable::VtblEntry::MetadataDropInPlace` entry, so add 1 to
+    // the allocation in case it doesn't.
+    let mut parts = Vec::with_capacity(methods.len() + 1);
```
spernsteiner, line 1132:
```suggestion
    // Reserve space for all methods plus the additional drop function.  Usually
    // `methods` will contain a few non-`VtblEntry::Method` entries, giving us extra
    // slots to work with, but we don't rely on that.
    let mut parts = Vec::with_capacity(methods.len() + 1);
```
I found the comment confusing (I had to check the loop body below to understand the connection to `VtblEntry::MetadataDropInPlace`) so I tried rewording it.

src/analyz/mod.rs
```diff
@@ -1088,7 +1126,26 @@ fn build_vtable_items<'tcx>(
...
+    // glue for trait objects, we always include a drop method in emitted
+    // vtables (and traits - see `emit_trait` above), and we always put it at
+    // the beginning of the method list so that `crucible-mir` can find and call
+    // it manually.
+    let concrete_ty = trait_ref.self_ty();
+    let drop_inst = ty::Instance::resolve_drop_in_place(tcx, concrete_ty);
+    let drop_def_id = drop_inst.def.def_id();
+    let drop_args = drop_inst.args;
+    // Note the `.insert(0, json)` in place of `.push(json)`
+    parts.insert(0, json!({
```
spernsteiner, line 1144: Why use `insert` here?  The vector is empty, so it's the same as a normal `push`.  I think it's fairly common to see code like
```Rust
let mut v = Vec::new();
v.push(fixed_item1);
v.push(fixed_item2);
for item in variable_items {
    v.push(f(item));
}
```

src/analyz/mod.rs
samcowger: Ok, that sounds reasonable. Does that mean we should change vtable serialization behavior to match this, such that it avoids serializing drop methods for trait objects without principal traits?

spernsteiner: For an unsupported case like this, I think it's fine to leave the two mismatched, as long as the inconsistency gets caught at some point in crucible-mir.  The main requirement for unsupported features is that it should either do the right thing or throw an error - it shouldn't silently do the wrong thing.

samcowger: Makes sense. At the moment, with the changes in https://github.com/GaloisInc/crucible/pull/1445, this would indeed trigger an error in `crucible-mir`, in the form of an index-out-of-range error when trying to access the first method of a vtable for a zero-method trait. The debug output associated with that error mentions that it occurred while `crucible` was trying to call a custom version of `drop_in_place`, so I hope that will be enough to steer a future victim of this issue in the right direction.

src/analyz/mod.rs
```diff
@@ -777,15 +777,62 @@ fn emit_trait<'tcx>(
...
+            "signature": drop_sig.to_json(ms),
+        }));
+    } else {
+        // I believe the only way this can happen is if this trait instance was
+        // derived from predicates (`TraitInst::from_dynamic_predicates`) with
+        // no principal trait, only marker traits. For example, `dyn Debug +
+        // Send + Sync`'s principal trait would be `Debug`, but `dyn Send +
+        // Sync` would have no principal trait, because `Send` and `Sync` are
+        // marker traits.
+        eprintln!("warning: no trait ref for {:?}?", ti);
```
spernsteiner, line 822: Should note explicitly that we decided not to support this for now, since we don't expect to see it in real code (because it doesn't seem useful)

spernsteiner (approved):

src/analyz/mod.rs
samcowger: Done.

PR #144, opened 2025-07-17
Add function for declaring uninterpreted functions

yav:
Adds a new function, which is overriden to declare that a symbol should be kept uninterpreted.

### DIFF
--- a/libs/crucible/cryptol.rs
+++ b/libs/crucible/cryptol.rs
@@ -18,6 +18,11 @@ pub fn override_<F>(f: F, module_path: &str, name: &str) {
     unimplemented!("cryptol::override")
 }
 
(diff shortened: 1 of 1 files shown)


libs/crucible/cryptol.rs
```diff
@@ -18,6 +18,11 @@ pub fn override_<F>(f: F, module_path: &str, name: &str) {
     unimplemented!("cryptol::override")
 }
 
+/// Mark the given name to be treated as uninterpreted for the duration
+/// of the current test.
```
spernsteiner, line 22: Please note in the doc comment that this applies only within Cryptol code that you call during the current test - it doesn't affect normal Rust functions.

spernsteiner (approved):

PR #145, opened 2025-07-17
analyz: `FnPtrShim`s have `RustCall` ABI

samcowger:
This ought to bring these shims in line with how Rust defines them - see e.g. [`Fn::call`](https://doc.rust-lang.org/std/ops/trait.Fn.html#tymethod.call), which is specified as `extern "rust-call"` (as are its `FnMut` and `FnOnce` siblings).


spernsteiner (approved):
Looks good, but please don't merge this until you have the corresponding Crucible change ready to go.  We've had trouble in the past with mir-json getting ahead of crucible-mir and causing confusion.

PR #146, opened 2025-07-18
Implement env_logger for debug prints

podhrmic:
Uses `env_logger` for debug printing. Has the unfortunate effect that *no info* is printed, suggesting that something else needs to happen. Sharing for awareness.

### DIFF
--- a/doc/rustc.md
+++ b/doc/rustc.md
@@ -236,3 +236,30 @@ stderr.
 Even with `--generate`, `mir-json-translate-libs` **will still run various
 `cargo` commands, including builds**, and temporarily create files, in order to
 determine which `mir-json` commands it should generate.
+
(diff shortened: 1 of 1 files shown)


podhrmic:
@spernsteiner @RyanGlScott this now works as expected.

RyanGlScott (commented):
Awesome, thanks!

The code itself looks fine to me, but I think we should document how to opt into higher verbosity levels. My understanding is that this is done by defining the `RUST_LOG` environment variable as needed—is that right? If so, it would be worth documenting this somewhere around [here](https://github.com/GaloisInc/mir-json/blob/774c8c3980cf868b5b10cc4808d966f36046b158/doc/rustc.md#L98-L104).

PR #149, opened 2025-07-25
libs: define overridable hooks for `core::slice::from_{ref,mut}`

samcowger:
This is similar to the approach in #147, which allowed hooking `core::array::from_ref`. Truly mimicking that approach would mean working with `core::array::from_mut`, which `slice::from_mut` calls. However, `crucible`'s special array representation makes it impossible to implement an override that would alias `array::from_mut`'s input and output references, and because those references are mutable, that shortcoming would be especially noticeable.


spernsteiner (approved):
Would it make sense to do the same thing for `slice::from_ref` while you're at it?

PR #152, opened 2025-08-06
Record layout information in types table

qsctr:
This bumps the schema version to 3.

### DIFF
--- a/src/analyz/to_json.rs
+++ b/src/analyz/to_json.rs
@@ -288,11 +288,18 @@ impl<'tcx> TyIntern<'tcx> {
         self.map.get(&ty).map(|x| x as &str)
     }
 
-    pub fn insert(&mut self, ty: ty::Ty<'tcx>, j: serde_json::Value) -> String {
-        let id = ty_unique_id(ty, &j);
+    /// `layout_j` should be [None] if the type is unsized.
+    pub fn insert(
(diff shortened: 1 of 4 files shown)


src/analyz/to_json.rs
```diff
@@ -288,11 +288,18 @@ impl<'tcx> TyIntern<'tcx> {
...
+        ty: ty::Ty<'tcx>,
+        ty_j: serde_json::Value,
+        layout_j: Option<serde_json::Value>,
+    ) -> String {
+        let id = ty_unique_id(ty, &ty_j);
         self.new_vals.push(json!({
             "name": &id,
-            "ty": j,
+            "ty": ty_j,
+            "layout": layout_j,
```
RyanGlScott, line 302: To me, `layout` feels like a part of the type-related information, so I wonder if this should go under `"ty"`.

RyanGlScott: Fair enough, I'm convinced.

RyanGlScott (approved):

src/analyz/to_json.rs
qsctr: Yeah, on the `crucible-mir` end the layouts will go into a separate `Map Ty (Maybe Layout)` in the `Collection`.

Issue #156, opened 2025-08-20
Support deriving `Symbolic` impls


RyanGlScott:
Kani supports the ability to derive impls for the [`Arbitrary` trait](https://model-checking.github.io/kani/reference/arbitrary.html), and it would be nice to be able to do the same with the `crucible` crate's [`Symbolic` trait](https://github.com/GaloisInc/mir-json/blob/753fe4b97b59035adc204aa426a3ae44b44822b9/libs/crucible/symbolic.rs#L5-L17). In principle, we should be able to do so by defining a [procedural derive macro](https://doc.rust-lang.org/reference/procedural-macros.html#derive-macros) for `Symbolic`. (See [here](https://github.com/model-checking/kani/blob/f27222d9228d10dfc3eee34fd02be1cbc20d2c39/library/kani_macros/src/lib.rs#L205-L209) for how such a macro is defined for Kani's `Arbitrary`.) The general approach would be:

1. Call `Symbolic::slice` to create each field in a struct or enum variant.
2. To pick between different variants in an enum, do something similar to what is documented [here](https://github.com/GaloisInc/crucible/tree/19dd3a402b0fe7f9ca75ede0c6361d82c93728f6/crux-mir#implement-symbolic-trait-for-a-custom-type).

PR #159, opened 2025-08-25
Support deriving Symbolic trait implementations

glguy:
See #156

(Please look closely at the way this proc_macro works, it's the first I've written. If something seems funny it's not likely that I'm doing something intentionally clever)

### DIFF
--- a/libs/crucible_proc_macros/src/lib.rs
+++ b/libs/crucible_proc_macros/src/lib.rs
@@ -7,7 +7,7 @@ use std::mem;
 use syn::fold::{self, Fold};
 use syn::spanned::Spanned;
 use syn::{parse_macro_input, parse_quote};
-use syn::{Attribute, Data, DeriveInput, Expr, ExprBlock, ExprCall, ExprMacro, Fields, Ident, Item, ItemFn, Path};
+use syn::{Attribute, Data, DeriveInput, Expr, ExprBlock, ExprCall, ExprMacro, Fields, Generics, GenericParam, Ident, Item, ItemFn, Path};
 
 #[derive(Clone)]
 struct Folder {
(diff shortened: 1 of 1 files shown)


libs/crucible_proc_macros/src/lib.rs
```diff
@@ -321,3 +334,89 @@ pub fn crux_spec_for(args: TokenStream, input: TokenStream) -> TokenStream {
...
+            });
+
+            let n = variants.len();
+            quote! {
+                {
+                    let variant: usize = crucible::Symbolic::symbolic_where("variant", |&i| i < #n);
+                    match variant {
+                        #( #arms, )*
+                        _ => crucible::crucible_assume_unreachable!(),
+                    }
```
spernsteiner, line 402: Currently the `i < n` constraint is being assumed in two places, once in `symbolic_where` and again through the use of `crucible_assume_unreachable!`.  Can you remove one or the other, so they can't get out of sync?  For the second one, I think there's a `crucible_assert_unreachable!` you could replace it with.

libs/crucible_proc_macros/src/lib.rs
```diff
@@ -321,3 +334,89 @@ pub fn crux_spec_for(args: TokenStream, input: TokenStream) -> TokenStream {
...
+            }
+        }
+
+        Data::Union(_) =>
+            panic!("Unions are not supported by derive(Symbolic)"),
+    };
+
+    let tokens = quote! {
+        #[cfg(crux)]
+        impl crucible::Symbolic for #name {
```
spernsteiner, line 413: Might be worth documenting that this derive doesn't support generics for now.  (Or else add support - IIRC there are some helpers in `syn` that make it pretty convenient.)

libs/crucible_proc_macros/src/lib.rs
```diff
@@ -52,33 +52,42 @@ impl Folder {
...
-            return report_error(call.span(), format_args!(
-                "found multiple calls to {}", self.subject_function_path.to_token_stream(),
-            ));
+            return report_error(
+                call.span(),
+                format_args!(
+                    "found multiple calls to {}",
+                    self.subject_function_path.to_token_stream(),
+                ),
+            );
```
spernsteiner, line 69: Please revert the whitespace changes on `Folder` - rustfmt has expanded a bunch of things that don't particularly need expanding

spernsteiner (approved):

Issue #131, opened 2025-06-27
Add regression tests


RyanGlScott:
Currently, `mir-json` itself doesn't have much in the way of tests aside from what is tested separately in `crux-mir` (in an entirely different repo). Given the existence of bugs like #128 and #129, it would be helpful to have regression tests to ensure that `mir-json` continues to compile these programs successfully without crashing. We should set up a regression test suite that we can run as part of `mir-json`'s CI.

Issue #129, opened 2025-06-26
Internal error when building code with nested static items


podhrmic:
Steps to reproduce:

1. `git clone git@github.com:chronotope/chrono.git`
2. `cd chrono`
3. `cargo build` (that proceeds normally)
4. `cargo crux-test --lib`
The relevant error output:

```
thread 'rustc' panicked at compiler/rustc_middle/src/hir/map/mod.rs:952:32:
internal error: entered unreachable code
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: could not compile `chrono` (lib test)

Caused by:
[...]

PR #165, opened 2025-09-02
Add Regression Testing to Build Workflow

sand7000:
This PR enhances the Docker build workflow by adding automated regression testing before pushing images to the registry. This addresses issue https://github.com/GaloisInc/mir-json/issues/131

## Workflow Improvements
- **New Job: `tag-x86-image`** - Creates a temporary test tag for the x86 image to enable testing
- **New Job: `regression-tests`** - Runs the test suite inside the Docker container before any images are pushed
- **Updated Trigger** - Changed from building only on master branch to building on all branches. The final push step still only occurs on master branch and only pushes a new nightly image if the regression tests pass.

## Test Framework Features
- Tests run against the built x86 Docker image in a containerized environment
- Test discovery from `tests/issues/` and `tests/regression/` directories
- Support for disabled tests via `disabled_tests.txt`
- Added example test for https://github.com/GaloisInc/mir-json/issues/128

## Benefits
The workflow now ensures that all images pass regression tests before being tagged and pushed to the registry.

## Open Questions
- **Nightly build behavior**: In this version the nightly build only pushes if the regressions pass. Would it be better to push regardless of test results to ensure we always have a nightly release?
- **ARM testing**: I don't think public repos have access to ARM CI runners, so I'm only testing on the x86 Docker image. Is this sufficient for now?

## Next Steps
- Add additional regression tests

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -125,3 +125,35 @@ jobs:
       - name: Inspect image
         run: |
           docker buildx imagetools inspect ${{ env.DOCKER_IMAGE }}:${{ steps.meta.outputs.version }}
+
+  run-regression-tests:
+    runs-on: ${{ matrix.runner }}
+    needs: [build-image]
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x86_64
+            runner: ubuntu-22.04
+          - arch: arm64
+            runner: ubuntu-22.04-arm
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Download digest
+        uses: actions/download-artifact@v4
+        with:
+          name: digests-${{ matrix.arch }}
+          path: digests
+
+      - name: Extract digest
+        id: digest
+        run: echo "digest=sha256:$(ls digests | head -n1)" >> $GITHUB_OUTPUT
+
+      - name: Run regression tests in Docker
+        run: |
+          docker run --rm \
+            -v "$PWD":"$PWD" \
+            -w "$PWD" \
+            ghcr.io/galoisinc/mir-json@${{ steps.digest.outputs.digest }} \
+            ./mir-json/tests/run-all.sh
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -45,7 +45,7 @@ jobs:
             -v "$PWD":"$PWD" \
             -w "$PWD" \
             local-test-image -c \
-            'ls -R'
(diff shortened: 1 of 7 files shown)


RyanGlScott:
Thanks for working on this, @sand7000!

Before responding to your open questions, some high-level suggestions:

* Currently, this PR only runs regression tests in the Docker CI jobs. This is a good start, but ultimately, I think we also want to run the tests on the CI jobs that run in pull requests as well (i.e., the jobs started by [this workflow](https://github.com/GaloisInc/mir-json/blob/0df7f01d26c60ea62fc1880d54dae4d4455dce48/.github/workflows/ci.yml)). This is especially important because the Docker CI jobs are only run nightly, so if a pull request introduces a test suite failure, then we wouldn't know about it for a while unless the pull request's CI catches it!
* As I understand it, this PR works by building the Docker images first, then creating a dedicated container that runs the tests. I think I would prefer a different approach where we instead run the tests as part of building the Docker image itself. This has a couple of advantages:
  1. Anyone who builds the Docker image locally can have some confidence that everything is working as expected, as they know that the tests pass on their locally built image.
  2. You would be able to greatly simplify the changes needed in the Docker CI workflow configuration, as you wouldn't need to start a separate container for running tests.

> **Nightly build behavior**: In this version the nightly build only pushes if the regressions pass. Would it be better to push regardless of test results to ensure we always have a nightly release?

I think we should only make the overall job pass if the tests pass as well. If there are test suite failures, then something terribly wrong has happened, and we should investigate it.

> **ARM testing**: I don't think public repos have access to ARM CI runners

Not true! In fact, `mir-json` (a public GitHub repo) is using an ARM-based `ubuntu-22.04-arm` CI runner (see [here](https://github.com/GaloisInc/mir-json/blob/0df7f01d26c60ea62fc1880d54dae4d4455dce48/.github/workflows/docker.yml#L21-L22)).

> [...] so I'm only testing on the x86 Docker image. Is this sufficient for now?

I think we should aspire to have equal testing coverage for both x86-64 and ARM. Luckily, if we move running the tests to be part of building the Docker image itself, then I think we would essentially get multi-architecture compatibility for free (or for very cheap, at least).

sand7000:
I added a test for [129](https://github.com/GaloisInc/mir-json/issues/129) and an another example test that checks the json output of a simple rs file. The test for [129](https://github.com/GaloisInc/mir-json/issues/129) fails of course because it is still in progress which means the build fails too as expected.

If [129](https://github.com/GaloisInc/mir-json/issues/129) is not high priority I could add the test to disabled_tests.txt to allow builds to complete.

PR #182, opened 2025-10-24
Add a JSON Schema spec for the MIR JSON format (version 5)

yav:


RyanGlScott (commented):
This looks cool! I have some questions about how this works, and it may be worth documenting the answers to this question in a `README` somewhere:

* This is located under `doc/schema-v4`. Is the intent here that we will create a copy of this folder every time that we perform a major version bump of the schema?
* How should this be updated whenever we do need to change it? Does it suffice to edit the `*.json` files directly, or is it better to use some kind of external tooling to do this?
* Is it easy to validate that a `mir-json`-generated JSON file matches this schema? If so, how does one do this?

yav:
We should probably decide on these things, in particular:
  * We could choose to either have 1 schema, that has conditional checks dependent on version, or complete copies for each new version.   JSON Schema is flexible enough to support a single spec, but things might get complicated (they already are), so it might be easier to just copy the whole thing for each new version, and then modify it.
  * You just need to modify the `*.json` files.   Having written all this, I think JSON Schema is not a good specification language as it is extremely verbose.  I tried looking for other languages but couldn't find anything reasonable (it was not a very detailed search admitedly).   I was considering making a little tool at some point that would allow you to write the spec in a more readable way, and generate the JSON Schema from that, but that's a pretty low priority I think.
  * I just used VSCode to test the schema on existing files, but there are [many tools for working with JSON Schemas](https://json-schema.org/tools?query=&sortBy=name&sortOrder=ascending&groupBy=toolingTypes&licenses=&languages=&drafts=&toolingTypes=&environments=&showObsolete=false&supportsBowtie=false). Not sure which ones are good though.

doc/mir-json-schema.ts
```diff
@@ -0,0 +1,407 @@
+/// The MIR JSON format
```
RyanGlScott: To accompany these `.ts` files, it would be helpful to have a `doc/README.md` file saying what these `*.ts` files, linking to the `simple-json-schema` repo as a resource for learning more. I think it would be helpful to link `doc/README.md` from the top-level README as well (probably in the [JSON schema section](https://github.com/GaloisInc/mir-json/blob/c9278101276acf40c191a28dd8bdbb9cbf08be86/README.md#json-schema)).

RyanGlScott (commented):
LGTM, aside from some documentation-related suggestions.

.github/ci.sh
```diff
@@ -16,6 +16,9 @@ zip_dist() {
   tar -czf "$name".tar.gz "$name"
 }
 
+# Download `simple-json-schema` and run it on mir-json files generated by 
+# the tests.   Note that, because of this, this should be run *after*
+# running the tests, so that we gener
```
RyanGlScott: Incomplete sentence?

RyanGlScott (approved):

PR #183, opened 2025-11-04
analyz: add `needs_drop` field to types in emitted JSON

samcowger:
This will help downstream consumers support Rust's [`needs_drop`](https://doc.rust-lang.org/std/intrinsics/fn.needs_drop.html) intrinsic. Most immediately, it will help `crucible-mir` do so - see https://github.com/GaloisInc/crucible/issues/1611.


RyanGlScott (approved):
Since there is some urgency on this, I talked with @aschwerdfeger-galois about this topic, and he agreed that we could cut release branches for Crucible _et al._ that do not include `needs_drop`-related changes. As such, I would be OK with landing this now.

PR #190, opened 2025-11-15
Add mir-json/schema version flag to most exes.

### LINKED ISSUE #77: --version should also print the mir-json version
Currently if you ask the mir-json tools for their version they print the rustc version, which is important, but realistically they should also print the mir-json version. Maybe also the current JSON schema version too, now that we have one.

In order to do this we need a version, but there is actually one in the cargo metadata (maybe we should bump it for the recent JSON changes?) so this shouldn't be that difficult to arrange.

chathhorn-galois:
Example output:
```
% saw-rustc --version                               
mir-json 0.1.0 (JSON schema version 5)
Rust toolchain nightly-2025-02-16
test build - extract output path - ["rustc", "--version", "--test", "--target", "aarch64-apple-darwin", "--cfg", "crux", "-Z", "ub-checks=false", "-L", "/Users/chathhorn/mir-json/rlibs", "--sysroot", "/Users/chathhorn/mir-json/rlibs"]
rustc 1.86.0-nightly (9cd60bd2c 2025-02-15)

% crux-rustc --version
mir-json 0.1.0 (JSON schema version 5)
Rust toolchain nightly-2025-02-16
normal build - ["rustc", "--version", "--target", "aarch64-apple-darwin", "--cfg", "crux", "-Z", "ub-checks=false", "-L", "/Users/chathhorn/mir-json/rlibs", "--sysroot", "/Users/chathhorn/mir-json/rlibs"]
rustc 1.86.0-nightly (9cd60bd2c 2025-02-15)

% mir-json --version
mir-json 0.1.0 (JSON schema version 5)
Rust toolchain nightly-2025-02-16
rustc 1.86.0-nightly (9cd60bd2c 2025-02-15)
```

Fixes #77.

### DIFF
--- a/src/bin/mir-json-translate-libs.rs
+++ b/src/bin/mir-json-translate-libs.rs
@@ -407,7 +407,6 @@ impl fmt::Display for CmdInvocation {
 }
 
 fn main() {
-
     let cwd: Utf8PathBuf = env::current_dir()
         .expect("cwd should be accessible")
         .try_into()
@@ -457,7 +456,6 @@ fn main() {
                 ),
         ])
         .get_matches();
-
     let custom_sources_dir = match arg_matches.value_of("libs") {
         Some(path) => fs::canonicalize(path)
             .expect("libs should be a valid path")
--- a/src/version.rs
+++ b/src/version.rs
@@ -3,8 +3,7 @@ use crate::schema_ver::SCHEMA_VER;
 use std::env::{self, Args};
 
 /// Check if the arguments contain version flags ("--version" or "-V").
-pub fn has_flag(args: &mut Args) -> bool
-{
+pub fn has_flag(args: &mut Args) -> bool {
     args.any(|arg| {
         arg == "--version" ||
         // Combined short flags like -vV


src/schema_ver.rs
```diff
@@ -1,9 +0,0 @@
-/// The version of the JSON schema that `mir-json` follows. This is intended for
-/// use by downstream tools to quickly determine if they are ingesting a MIR
-/// JSON file that is compatible with the version of the schema that they are
-/// expecting.
-///
-/// Each version of the schema is assumed to be backwards-incompatible with
-/// previous versions of the schema. As such, any time this version number is
-/// bumped, it should be treated as a major version bump.
-pub const SCHEMA_VER: u64 = 5;
```
RyanGlScott: Why did you need to move the schema version from this `src/schema_ver.rs` file?

This is a somewhat practical concern, as the [`mir-json` CI](https://github.com/GaloisInc/mir-json/blob/bf7285d76f3798fc373372eb9aa9640524b74873/.github/workflows/docker.yml#L86) currently assumes that the schema version is located in a file with that name.

src/bin/cargo-mir-json.rs
```diff
@@ -23,18 +26,14 @@ fn show_help() {
...
 fn main() {
     // Check for version and help flags even when invoked as 'cargo-mir-json'
     if env::args().any(|a| a == "--help" || a == "-h") {
         show_help();
         return;
     }
-    if env::args().any(|a| a == "--version" || a == "-V") {
-        show_version();
+    if version::has_flag(&mut env::args()) {
+        println!("{}", version::string());
```
RyanGlScott: Since this `println!` invocation is repeated a lot, it may be worth factoring this out into its own `show_version()` version, similar to the already-existing `show_help()` function (as seen above).

src/schema_ver.rs
RyanGlScott: Yes, I'd favor keeping them apart. These concerns are somewhat separate: `schema_ver.rs` exists to track solely the schema version, while `version.rs` handles version types beyond just the schema version.

PR #202, opened 2025-12-02
Print `rustc` version information before `mir-json` version

### LINKED ISSUE #201: `mir-json` fails to compile `libc` crate after #190
After https://github.com/GaloisInc/mir-json/pull/190, `mir-json` is no longer able to build the `libc` crate using `cargo crux-test` or `cargo saw-build`. To reproduce this issue, create a new Cargo project with the following `Cargo.toml` file:

```toml
[package]
name = "libc-example"
version = "0.1.0"
edition = "2024"

[dependencies]
libc = "0.2.178"
```

If you compile it with `cargo saw-build`, it will fail thusly:

```
$ cargo saw-build
<snip>
   Compiling libc v0.2.178
error: failed to run custom build command for `libc v0.2.178`

Caused by:
  process didn't exit successfully: `/Users/rscott/Documents/Hacking/Rust/libc-example/target/debug/build/libc-61997d9374391303/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=build.rs

  --- stderr

  thread 'main' (23069704) panicked at /Users/rscott/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.178/build.rs:224:5:
  assertion `left == right` failed: Failed to get rustc version
    left: Some("mir-json 0")
   right: Some("rustc 1")
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The problem is that `libc` has a [custom build script](https://github.com/rust-lang/libc/blob/46845987076649c0458810e323162d5c8b40b61e/build.rs#L202-L244) that queries the output of `rustc --version` and then parses it. This build script is very sensitive to the format of this output. Before https://github.com/GaloisInc/mir-json/pull/190, the output of `rustc --version` (where `rustc` has been overridden with `mir-json`'s version of `rustc`) would look something like this:

```
rustc 1.91.0-nightly (02c7b1a7a 2025-09-13)
```

After https://github.com/GaloisInc/mir-json/pull/190, however, it now looks like:

```
mir-json 0.1.0 (JSON schema version 6)
Rust toolchain nightly-2025-09-14
rustc 1.91.0-nightly (02c7b1a7a 2025-09-13)
```

I propose that we move the `rustc 1.91.0-nightly (02c7b1a7a 2025-09-13)` bit back to the first line such that `libc`'s build script will continue to work.

(Note that although we build the `libc` crate as part of the modified Rust standard libraries that `mir-json` bundles, `mir-json-translate-libs` doesn't override `rustc` during this build process, which is why this oversight wasn't caught by `mir-json`'s CI.)

RyanGlScott:
This is important to be able to build crates such as `libc`.

Fixes #201.

### DIFF
--- a/src/bin/mir-json-rustc-wrapper.rs
+++ b/src/bin/mir-json-rustc-wrapper.rs
@@ -22,7 +22,7 @@ use mir_json::link;
 use mir_json::version;
 use rustc_ast::Crate;
 use rustc_middle::ty::TyCtxt;
-use rustc_driver::Compilation;
+use rustc_driver::{Callbacks, Compilation};
 use rustc_interface::interface::{Compiler, Config};
 use rustc_session::config::{Externs, ExternLocation, OutFileName, OutputFilenames};
 use rustc_span::Symbol;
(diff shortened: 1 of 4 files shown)


chathhorn-galois (approved):

src/bin/mir-json-rustc-wrapper.rs
```diff
@@ -166,10 +180,6 @@ fn go() -> ExitCode {
     // First arg is the name of the `rustc` binary that cargo means to invoke, which we ignore.
     let mut args: Vec<String> = std::env::args().skip(1).collect();
```
chathhorn-galois: If `args[0]` is supposed to be `rustc` invoked below, it looks like it's skipped here. Not sure why this was apparently passing before though.

Issue #153, opened 2025-08-18
Make `crucible_*_hook` persist in MIR code even when optimizations are applied


RyanGlScott:
Recently, I discovered code that looks roughly like this:

```rs
// test.rs
pub fn cast(s: &[u8]) -> &[u8; 8] {
    s.try_into().unwrap()
}
```

A peculiar property of this code is that SAW is able to verify it without optimizations applied, but it is not able to verify it when optimizations are applied. More precisely, if you have the following SAW specification for `cast`:

```
// test.saw
enable_experimental;

let cast_spec = do {
    s <- mir_fresh_var "s" (mir_array 8 mir_u8);
    s_ref <- mir_ref_of (mir_term s);
    let s_slice = mir_slice_value s_ref;

    mir_execute_func [s_slice];
};

m <- mir_load_module "test.linked-mir.json";
mir_verify m "test::cast" [] false cast_spec z3;
```

Then SAW will succeed if you compile the Rust code with no optimizations applied (the default `rustc` settings):

```
$ saw-rustc --edition=2021 -Copt-level=0 test.rs
<snip>
$ saw test.saw
[11:26:52.084] Loading file "test.saw"
[11:26:52.310] Verifying test/d305a93c::cast[0] ...
[11:26:52.316] Simulating test/d305a93c::cast[0] ...
[11:26:52.316] Checking proof obligations test/d305a93c::cast[0] ...
[11:26:52.317] Proof succeeded! test/d305a93c::cast[0]
```

On the other hand, SAW will fail if you compile the Rust code with higher optimization levels:

```
$ saw-rustc --edition=2021 -Copt-level=3 test.rs
<snip>
$ saw test.saw
[11:27:38.928] Loading file "test.saw"
[11:27:39.148] Verifying test/7cf03e97::cast[0] ...
[11:27:39.154] Simulating test/7cf03e97::cast[0] ...
[11:27:39.154] Stack trace:
   (builtin) in mir_verify
   test.saw:13:1-13:48 (at top level)
Symbolic execution failed.
Abort due to assertion failure:
  ./libs/core/src/slice/mod.rs:869:27: 869:58: error: in test/7cf03e97::cast[0]
  Translation error in test/7cf03e97::cast[0]: unimplemented cast: Misc
    ty: TyRawPtr (TySlice (TyUint B8)) Immut
    as: TyRawPtr (TyArray (TyUint B8) 8) Immut
```

Here is what is going on:

[...]

PR #203, opened 2025-12-02
Reapply "Mark Crucible-related hook functions as `#[inline(never)]`"

RyanGlScott:
This re-applies the changes from #154, which were omitted from #200 due to an oversight. It is a bit unclear whether or not `#[inline(never)]` is strictly necessary here (see https://github.com/GaloisInc/mir-json/issues/153#issuecomment-3602259898), but better to be on the safe side.

To make it more likely that we will remember to apply this change in the future, I have started a "Notes" section in `libs/Patches.md` that very explicitly calls out the needs for `#[inline(never)]`, and I have cited this note from the relevant patch descriptions. Going forward, we can use this "Notes" section to describe other aspects of patches that require more in-depth explanations that what can be afforded in the brief patch descriptions.

### DIFF
--- a/libs/Patches.md
+++ b/libs/Patches.md
@@ -8,9 +8,16 @@ between Rust versions, so applying the patches is rarely as straightforward as
 running `git apply`.
 
 As a compromise, this document contains high-level descriptions of each type of
-patch that we apply, along with rationale for why the patch is necessary. The
-intent is that this document can be used in conjunction with `git blame` to
-identify all of the code that was changed in each patch.
+patch that we apply, along with the date in which it was last applied and the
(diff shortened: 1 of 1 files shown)


libs/Patches.md
```diff
@@ -212,3 +224,26 @@ identify all of the code that was changed in each patch.
 
   This feature was using an `AtomicPtr<()>` with a cast from function
   pointer to `*mut ()` that we don't support in its initializer.
+
+## Notes
+
+This section contains with more detailed noted about why certain patches are
```
samcowger:
```suggestion
This section contains more detailed noted about why certain patches are
```

samcowger (approved):

PR #205, opened 2025-12-02
implement cases for TyKind::CoroutineClosure

spernsteiner:
Judging by [this code][layout-of-cc] in the implementation of `layout_of`, `TyKind::CoroutineClosure` works basically the same as `TyKind::Closure`, so I've copied those cases and just changed the names.

[layout-of-cc]: https://github.com/rust-lang/rust/blob/646a3f8c15baefb98dc6e0c1c1ba3356db702d2a/compiler/rustc_ty_utils/src/layout.rs#L559

### DIFF
--- a/SCHEMA_CHANGELOG.md
+++ b/SCHEMA_CHANGELOG.md
@@ -3,6 +3,11 @@ The following document describes the changes to the JSON schema that
 as a changelog for the code in the `mir-json` tools themselves, which are
 versioned separately.)
 
+## 7
+
+Add new `InlineType`, `ConstVal`, and `AggregateKind` cases for
+`TyKind::CoroutineClosure`, mimicking the cases for `TyKind::Closure`.
+
 ## 6
 
 Upgrade the supported Rust toolchain to `nightly-2025-09-14` (`1.91`).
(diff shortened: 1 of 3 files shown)


src/analyz/ty_json.rs
```diff
@@ -397,15 +397,17 @@ impl<'tcx> ToJson<'tcx> for ty::Instance<'tcx> {
                 // with the cases in rustc's `build_clone_shim` function (see
                 // https://doc.rust-lang.org/1.86.0/nightly-rustc/src/rustc_mir_transform/shim.rs.html#432-455).
                 //
-                // TODO(#196): We are currently missing two cases for
-                // TyKind::Coroutine and TyKind::CoroutineClosure. We should
-                // add these cases, taking inspiration from how
+                // TODO(#196): We are currently missing a case for
+                // TyKind::Coroutine. We should
+                // add this cases, taking inspiration from how
```
RyanGlScott:
```suggestion
                // add this case, taking inspiration from how
```

RyanGlScott (approved):
Do you think it's worth bumping the schema version to reflect these changes?

src/analyz/ty_json.rs
spernsteiner: Fixed in bb9ba85b53d6da6ceffd0ba31f174b01820e0b67

PR #208, opened 2025-12-11
`libs`: Patch `MaybeUninit::assume_init()` to avoid unsupported pointer cast

RyanGlScott:
https://github.com/rust-lang/rust/commit/8362508989c8941ccbba26aaa8c787cc492c6fef (which debuted in Rust 1.90.0) changed the implementation of `MaybeUninit::assume_init()` to use a raw pointer cast for the sake of improving error messages in Miri. While good for Miri, this is bad for `crucible-mir`, as this pointer cast can involve distinct Crucible type representations, which is a type of cast that `crucible-mir` doesn't support. In particular, if using `array::from_fn()` to construct a length-0 array, then this would require casting from `UnitRepr` (due to the use of a zero-sized type) to a `MirAggregateRepr`, which `crucible-mir` does not currently support.

This commit reverts back to the old implementation of `assume_init()`, which does not have a problematic pointer cast.

Towards https://github.com/GaloisInc/crucible/issues/1674.

### DIFF
--- a/libs/Patches.md
+++ b/libs/Patches.md
@@ -197,6 +197,14 @@ into the main commit for that patch, and then the *Update* line can be removed.
   difficult because we don't have a good way to detect which union variant is
   active.
 
+* Remove raw pointer cast in `MaybeUninit::assume_init()` (last applied: December 11, 2025)
+
+  The implementation of `assume_init()` uses a raw pointer cast that Crucible
+  does not support in all cases. This is needed in particular to support
+  calling `array::from_fn()` to construct a length-0 array, as this would
+  otherwise require a cast from a `UnitRepr` (due to the use of a zero-sized
+  type) to a `MirAggregateRepr`, which Crucible does not support.
+
 * Use `crucible_array_from_ref_hook` in `core::array::from_ref` (last applied: November 25, 2025)
 
   The actual implementation uses a pointer cast that Crucible can't handle. See
--- a/libs/core/src/mem/maybe_uninit.rs
+++ b/libs/core/src/mem/maybe_uninit.rs
@@ -616,9 +616,7 @@ impl<T> MaybeUninit<T> {
         // This also means that `self` must be a `value` variant.
         unsafe {
             intrinsics::assert_inhabited::<T>();
-            // We do this via a raw ptr read instead of `ManuallyDrop::into_inner` so that there's
-            // no trace of `ManuallyDrop` in Miri's error messages here.
-            (&raw const self.value).cast::<T>().read()
+            ManuallyDrop::into_inner(self.value)
         }
     }
 


RyanGlScott:
> Why is `UnitRepr` appearing here?

To illustrate why this happens, here is a slightly smaller version of the program from https://github.com/GaloisInc/crucible/issues/1674 that fails in the same way:

```rs
use std::mem::MaybeUninit;

pub fn f<const N: usize>() -> [MaybeUninit<u8>; N] {
    let array: MaybeUninit<[MaybeUninit<u8>; N]> = MaybeUninit::uninit();
    unsafe { array.assume_init() }
}

#[crux::test]
pub fn g() -> [MaybeUninit<u8>; 0] {
    f()
}
```

Let's walk through the execution of `g()` step by step after calling into `f::<0>()`:

1. `MaybeUninit::uninit()` is invoked, which triggers [this custom override](https://github.com/GaloisInc/crucible/blob/594305a3726478752247e8489ce1dcbe69948346/crucible-mir/src/Mir/TransCustom.hs#L2093-L2101). This, in turn, creates an `initialValue` of type `MaybeUninit<[MaybeUninit<u8>; 0]>`.
2. There is a [special case in `initialValue`](https://github.com/GaloisInc/crucible/blob/594305a3726478752247e8489ce1dcbe69948346/crucible-mir/src/Mir/TransTy.hs#L1509-L1513) for ADTs that are marked `repr(transparent)`, such as `MaybeUninit`. This causes it to create an initial value for the underlying field type, as computed by `reprTransparentFieldTy`.
3. What _is_ the underlying field type for `MaybeUninit<[MaybeUninit<u8>; 0]>`, anyway? The definition of `MaybeUninit` is:

   ```rs
   #[repr(transparent)]
   union MaybeUninit<T> {
       uninit: (),
       value: ManuallyDrop<T>,
   }

   #[repr(transparent)]
   struct ManuallyDrop<T: ?Sized> {
       value: T,
   }
   ```

   As such, `MaybeUninit<[MaybeUninit<u8>; 0]>` has two fields: `uninit: ()` and `value: ManuallyDrop<[MaybeUninit<u8>; 0]>`. As it turns out, both of these fields are zero-sized! This means that `reprTransparentFieldTy` could reasonably return either field type.

   In practice, `reprTransparentFieldTy` disambiguates this by choosing [the first field](https://github.com/GaloisInc/crucible/blob/594305a3726478752247e8489ce1dcbe69948346/crucible-mir/src/Mir/TransTy.hs#L400) (`uninit: ()`). This means that the `array` variable will be given an initial value for type `()`, thereby giving it the Crucible type representation `UnitRepr`.
4. Note that things would be different if the size of the array were non-zero, as this would make `ManuallyDrop<[MaybeUninit<u8>; N]>` a non-zero-sized type. In that case, `reprTransparentFieldTy` knows that there is exact one non-zero-sized field and returns `value` instead of `uninit`, thereby giving the `array` variable the Crucible type representation for arrays (`MirAggregateRepr`).

samcowger:
As an alternative to making that reversion, would it make sense to change `crucible-mir` to make `()` use `MirAggregateRepr` rather than `UnitRepr`, do you think? Unit values would then be backed by a zero-size `MirAggregate`, which I imagine would solve https://github.com/GaloisInc/crucible/issues/1674 and possibly avoid others down the line.

RyanGlScott:
That's an intriguing idea for sure, although it would take a fair bit of work to implement it (there are quite a number of places where `UnitRepr` is hard-coded that would need to be audited). Perhaps it would be worthwhile, however, given that this sort of cast between zero-sized type representations is pretty common in practice.

If we did make this change, one interesting observation is that nearly every zero-sized type in Rust would now be represented with `MirAggregate`s in `crucible-mir`. I say "nearly" because the other exceptions to that rule are `TyFnDef`/`TyNever`, which are [also represented with `UnitRepr`](https://github.com/GaloisInc/crucible/blob/9fa8f120326f3d5accca726b9c35d225ba2bb0ca/crucible-mir/src/Mir/TransTy.hs#L262-L265)—perhaps we ought to consider changing that to `MirAggregateRepr` for consistency?

spernsteiner:
I'd be in favor of changing `()`, `TyFnDef`, and `TyNever` to all use `MirAggregateRepr` instead of `UnitRepr`.  Changing just `()` means we could hit the same issue later with `MaybeUninit<TyFnDef>` (which is weird on its own but could conceivably show up via generics).

PR #259, opened 2026-04-06
Add coroutine_field family of operations

glguy:


RyanGlScott:
This looks generally reasonable, although I think it would be helpful to add some docstrings to these functions to explain what they're used for (and perhaps to give an example of how they might be used). For instance, am I understanding correctly that `coroutine_field<_, _, I>(co)` will access the `I`th argument that was passed to the coroutine `co`?

Relatedly, is there a reason that `I` is passed as a const parameter instead of as, say, a `usize` value? I'm not sure of what the tradeoffs are for making this information static vs. dynamic.

spernsteiner:
> Relatedly, is there a reason that `I` is passed as a const parameter instead of as, say, a `usize` value? I'm not sure of what the tradeoffs are for making this information static vs. dynamic.

The easiest way to access a coroutine upvar is with the helper in `TransTy`, which takes an `Int` index and produces Crucible code in the `MirGenerator` monad.  Since it uses `MirGenerator`, it can only be called from a `TransCustom`-style override (not from `Mir.Intrinsics` or `Mir.Overrides`).  But `TransCustom` overrides can only see (at most, via `CustomMirOp`) the MIR syntax of their `Operand`s, which means for code like `&*coroutine_field::<T, U>(coroutine, i)` in `coroutine_field_ref`, the override will see the variable `i` (i.e. some kind of `Lvalue`) instead of the actual index as a `ConstInt`.

RyanGlScott (approved):

PR #261, opened 2026-04-10
Add macro for making async fn overrides easy

glguy:


glguy:
I'll update the crux-mir test to use this macro instead of the raw `coroutine_field_ref` once this is merged

libs/crucible/coroutine.rs
```diff
@@ -35,3 +39,149 @@ pub fn coroutine_field_ref<T, U, const I: usize>(coroutine: &T) -> &U {
...
+/// Trivial async executor for use during symbolic execution.
+/// All poll operations must immediately succeed when using this.
+pub fn trivial_block_on<F: Future>(mut fut: F) -> F::Output {
+    let mut cx = Context::from_waker(Waker::noop());
+    let mut fut = unsafe { Pin::new_unchecked(&mut fut) };
+    match fut.as_mut().poll(&mut cx) {
+        Poll::Ready(x) => x,
+        Poll::Pending => panic!("unexpected Pending"),
+    }
+}
```
RyanGlScott, line 200: What is the use case for this function? As far as I can tell, it's not used anywhere in the implementation of the macro, so it's unclear to me if this is meant to be used externally.

spernsteiner:
> I'll update the crux-mir test to use this macro instead of the raw `coroutine_field_ref` once this is merged

Please make a new test for that.  Then if something breaks during a future toolchain upgrade, it will be easier to tell whether `coroutine_field` is broken or whether the macro is broken.

libs/crucible/coroutine.rs
```diff
@@ -35,3 +39,149 @@ pub fn coroutine_field_ref<T, U, const I: usize>(coroutine: &T) -> &U {
...
+
+            // Compiler is too smart to just use panic!() to make the arguments
+            fn __undef<T>(_label: &str) -> T { unimplemented!() }
+
+            pub type __TargetFuture = impl core::future::Future<Output = $retTy>;
+
+            #[define_opaque(__TargetFuture)]
+            fn __define_target_future() -> __TargetFuture {
+                $target( $(__undef(stringify!($UpvarTy))),* )
+            }
```
RyanGlScott, line 159: What is this function used for?

spernsteiner: The goal here is to assign a name to the anonymous coroutine type returned by `$target` using `feature(type_alias_impl_trait)` so we can later obtain its `poll` method.  The type alias is defined above, and the purpose of this function is to constrain the hidden type of the `impl Trait` to be exactly the return type of `$target`.  This function only exists for the side effect it has on type checking; it's never actually called.

libs/crucible/coroutine.rs
glguy: It's a trivial execution environment that can be used while verifying the behavior of asynchronous functions

RyanGlScott: It wasn't super clear from reading this that this is meant to be used for verification purposes. Can you document this (perhaps giving an example of how you'd use this in a `#[crux::test]` setting)?

libs/crucible/coroutine.rs
RyanGlScott: Ah, that makes sense. Can you document this somewhere in a nearby comment?

libs/crucible/coroutine.rs
```diff
@@ -35,3 +39,161 @@ pub fn coroutine_field_ref<T, U, const I: usize>(coroutine: &T) -> &U {
...
+/// 
+/// # Example
+/// 
+/// ```ignore
+/// #[crux::test]
+/// fn example_test() {
+///     override_get_random();
+///     let result = trivial_block_on(async_fn_using_get_random());
+///     crucible_assert!(property of thing_using_get_random);
+/// }
```
RyanGlScott, line 191:
```suggestion
/// ```ignore
/// #[crux::test]
/// fn example_test() {
///     override_get_random();
///     let result = trivial_block_on(async_fn_using_get_random());
///     crucible_assert!(property of thing_using_get_random);
/// }
/// ```
```

RyanGlScott (approved):

Issue #193, opened 2025-11-21
Downgrade `eprintln!` calls in `mir-json` to `debug!` logs


RyanGlScott:
There are various places in `mir-json` that unconditionally dump fairly verbose text to stderr using the `eprintln!` macro. Some of the more annoying ones include these two in the `crucible_proc_macros` crate:

https://github.com/GaloisInc/mir-json/blob/3b2fafa36dbce2c1b527184ce7004c4ff73e383c/libs/crucible_proc_macros/src/lib.rs#L321
https://github.com/GaloisInc/mir-json/blob/3b2fafa36dbce2c1b527184ce7004c4ff73e383c/libs/crucible_proc_macros/src/lib.rs#L423

These cause the results of the `crux_spec_for` and `derive(Symbolic)` proc macros to be dumped in their entirety, e.g.,

```
output = #[crux :: test] fn f_equiv()
{
    crucible :: method_spec :: clobber_globals();
    {
        let arr = < [u8; 4] > :: symbolic("arr"); let vec = mk_vec(& arr); let
        output_impl = { f(& vec) }; crucible_assert!
        (output_impl == arr.into_iter().map(|n| n as u64).sum());
    }
} fn f_equiv_spec() -> crucible :: method_spec :: MethodSpec
{
    let mut __crux_msb : crucible :: method_spec :: MethodSpecBuilder;
    {
        let arr = < [u8; 4] > :: symbolic("arr"); let vec = mk_vec(& arr); let
        output_impl =
        {
            let __crux_func = f; __crux_msb = crucible :: method_spec ::
            MethodSpecBuilder :: new(__crux_func);
            __crux_msb.add_arg(& (& vec)); __crux_msb.gather_assumes(); let
            result = if false { __crux_func(& vec) } else
            { crucible :: Symbolic :: symbolic("f_result") };
            __crux_msb.set_return(& result); result
        }; crucible_assert!
        (output_impl == arr.into_iter().map(|n| n as u64).sum());
    } __crux_msb.gather_asserts(); __crux_msb.finish()
}
```

This adds a fair amount of clutter to stderr. We should downgrade this to `debug!` such that it is off by default (but still allowing users to opt in if they want).

There are also a variety of other `eprintln!` calls throughout the source code, which should also be put under similar scrutiny:

```
$ git grep -n "eprintln\!" src/
src/analyz/mod.rs:829:        eprintln!("warning: no trait ref for {:?}?", ti);
[...]

PR #263, opened 2026-04-16
`crucible_proc_macros`: Demote `eprintln!` calls to `debug!` logs

RyanGlScott:
This requires adding `env_logger` and `log` dependencies to the `crucible_proc_macros` crate, but these are extremely lightweight dependencies.

Addresses some (but not all) of #193.


samcowger (approved):
I tend to think we should leave #193 open after these changes, since it pertains to several other still-extant uses of `eprintln!` elsewhere in the tooling, but otherwise LGTM.

PR #264, opened 2026-04-22
`libs`: Special-case `slice::as(_mut)_array` for zero-length arrays

RyanGlScott:
We cannot call `slice::as_chunks_unchecked(_mut)` for zero-length arrays, so we must add special cases to handle these. See the newly added note in `libs/Patches.md` for the full rationale.

Towards https://github.com/GaloisInc/crucible/issues/1814.

### DIFF
--- a/libs/Patches.md
+++ b/libs/Patches.md
@@ -312,6 +312,12 @@ into the main commit for that patch, and then the *Update* line can be removed.
   `as_chunks_unchecked(_mut)`: `first_chunk(_mut)`, `split_first_chunk(_mut)`,
   `last_chunk(_mut)`, `split_last_chunk(_mut)`, `slice::as(_mut)_array`.
 
+  See also the "Limitations of zero-length arrays and
+  `slice::as_chunks_unchecked(_mut)`" note below.
+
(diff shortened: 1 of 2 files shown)


libs/Patches.md
```diff
@@ -312,6 +312,9 @@ into the main commit for that patch, and then the *Update* line can be removed.
   `as_chunks_unchecked(_mut)`: `first_chunk(_mut)`, `split_first_chunk(_mut)`,
   `last_chunk(_mut)`, `split_last_chunk(_mut)`, `slice::as(_mut)_array`.
 
+  See also the "Limitations of zero-length arrays and
+  `slice::as_chunks_unchecked(_mut)`" note below.
+
```
qsctr, line 317: Could you also add an *Update* line with a date, so that we can remember to make this change in a future toolchain upgrade?

spernsteiner (approved):
Looks good to me, aside from @qsctr's comment

libs/Patches.md
RyanGlScott: Done.

qsctr (approved):

PR #265, opened 2026-04-22
libs fixes for tokio

spernsteiner:
* Always use `std::sys::thread_local::destructors::list`, which is simpler than `destructors::linux_like`
* Replace all variants of `core::slice::memchr::memchr` with `memchr_naive`

The destructor change let me get a trivial tokio example running on Linux:
```Rust
pub async fn add_two_randoms() -> u32 {
    let x = get_random(1, 5).await;
    let y = get_random(2, 6).await;
    x + y
}

async fn get_random(lo: u32, hi: u32) -> u32 {
    lo
}

#[cfg(crux)]
mod verification {
    use super::*;

    /// Verify that add_two_randoms produces a number in range
    #[crux::test]
    fn add_two_randoms_spec() {
        let runtime = tokio::runtime::Builder::new_current_thread().build().unwrap();
        runtime.block_on(add_two_randoms());
        std::mem::forget(runtime);
    }
}
```

The `mem::forget` is needed for now since there's some thread park/unpark logic in the runtime's destructor that crux-mir can't handle.

The switch to `memchr_naive` fixes an issue that (IIRC) @glguy ran into while testing a similar example on MacOS.  I didn't see that particular error from tokio when I was testing on Linux, but I can confirm that the change causes `CStr::from_bytes_with_nul(b"long string bigger than two words\0")` to succeed where it previously failed.  (`len >= 2 * size_of::<usize>()` is the cutoff for switching to the optimized `memchr` implementation that was causing problems for crux-mir.)

### DIFF
--- a/libs/std/src/sys/thread_local/mod.rs
+++ b/libs/std/src/sys/thread_local/mod.rs
@@ -55,27 +55,9 @@ cfg_select! {
 /// single callback that runs all of the destructors in the list.
 #[cfg(all(target_thread_local, not(all(target_family = "wasm", not(target_feature = "atomics")))))]
 pub(crate) mod destructors {
-    cfg_select! {
-        any(
-            target_os = "linux",
-            target_os = "android",
-            target_os = "fuchsia",
-            target_os = "redox",
-            target_os = "hurd",
-            target_os = "netbsd",
-            target_os = "dragonfly"
-        ) => {
-            mod linux_like;
-            mod list;
-            pub(super) use linux_like::register;
-            pub(super) use list::run;
-        }
-        _ => {
-            mod list;
-            pub(super) use list::register;
-            pub(crate) use list::run;
-        }
-    }
+    mod list;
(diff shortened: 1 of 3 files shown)


libs/std/src/sys/thread_local/mod.rs
```diff
@@ -56,20 +56,6 @@ cfg_select! {
 #[cfg(all(target_thread_local, not(all(target_family = "wasm", not(target_feature = "atomics")))))]
 pub(crate) mod destructors {
     cfg_select! {
```
RyanGlScott: I suppose this `cfg_select!` is now redundant, given that there is only a single case left?

RyanGlScott (approved):

libs/std/src/sys/thread_local/mod.rs
spernsteiner: Good point—I've removed it entirely

PR #267, opened 2026-04-23
add size and align to vtables (schema ver 11)

spernsteiner:
Adds `size` and `align` fields to vtable definitions.  These fields are non-optional, since only sized types can be coerced to `&dyn Trait`.  These will initially be used in crucible-mir to implement the `dyn` cases of `size_of_val` and `align_of_val`.

Longer term, the motivation for adding this is that the vtable alignment is needed when representing structs with `MirAggregateRepr`.  In `CustomDST<dyn Trait>`, the offset of the last field depends on its alignment, and its alignment is always the alignment of the erased type.  So computing the offset of the last field requires fetching the alignment from the vtable.

Currently the output JSON doesn't contain any link from a vtable to its `Self` type, since nothing needs it currently.  It seemed easier to add the size and align fields directly to the vtable, rather than adding a reference to a `Ty` that we otherwise wouldn't use.

### DIFF
--- a/SCHEMA_CHANGELOG.md
+++ b/SCHEMA_CHANGELOG.md
@@ -3,6 +3,10 @@ The following document describes the changes to the JSON schema that
 as a changelog for the code in the `mir-json` tools themselves, which are
 versioned separately.)
 
+## 11
+
+Add `size` and `align` fields to `dyn Trait` vtables.
+
 ## 10
 
 Add field offsets to the layout information for struct-like types.  This is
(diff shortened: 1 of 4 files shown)


spernsteiner:
The Crucible side of this is still WIP but should be ready soon-ish

RyanGlScott (commented):
Make sure to update this part of the JSON schema:

https://github.com/GaloisInc/mir-json/blob/2fc80d73eaede1e394a4ec6970fb5ca9e33efd73/doc/mir-json-schema.ts#L198-L202

RyanGlScott (approved):

PR #268, opened 2026-05-01
emit layout info for unsized types

spernsteiner:
This is needed for handling the unsized / custom DST cases when migrating structs to `MirAggregateRepr`.

It's safe to merge this on its own, since the new information it adds will simply be ignored by older crucible-mir.

Fixes https://github.com/GaloisInc/mir-json/issues/257.

### DIFF
--- a/src/analyz/ty_json.rs
+++ b/src/analyz/ty_json.rs
@@ -665,16 +665,15 @@ impl<'tcx> ToJson<'tcx> for ty::Ty<'tcx> {
                         Some(offsets.iter().map(|&off| off.bytes()).collect::<Vec<_>>())
                     },
                 };
-                if layout.is_sized() {
-                    Some(json!({
-                        "align": layout.align.abi.bytes(),
-                        "size": layout.size.bytes(),
-                        "field_offsets": field_offsets
-                    }))
-                } else {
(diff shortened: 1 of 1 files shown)


src/analyz/ty_json.rs
```diff
@@ -665,16 +665,15 @@ impl<'tcx> ToJson<'tcx> for ty::Ty<'tcx> {
...
-                        "field_offsets": field_offsets
-                    }))
-                } else {
-                    // TODO: provide field offsets even for unsized types
-                    None
-                }
+                // Output size, align, and field_offsets for both sized and unsized types.  For
+                // unsized types, this is the minimum size and alignment: `dyn Trait` has size 0
+                // and align 1, and `[T]` has size 0 and `align_of::<T>()`.  Layouts for unsized
+                // structs / custom DSTs are computed accordingly.
```
RyanGlScott: Is it worth clarifying here what `field_offsets` represents for unsized types?

RyanGlScott (approved):

src/analyz/ty_json.rs
spernsteiner: Expanded on this in 22cdd743

PR #271, opened 2026-05-26
Port mir-json binaries to Windows

AmeliaRose802:
Ports the `mir-json` toolchain binaries (`crux-rustc`, `mir-json-rustc-wrapper`, `saw-rustc`, `mir-json-translate-libs`) and `build.rs` to work on Windows hosts. Tested with `rustc` nightly + MSVC toolchain.

3 commits:
1. **Port mir-json binaries to Windows** — `src/bin/crux-rustc.rs`, `mir-json-rustc-wrapper.rs`, `saw-rustc.rs`: handle Windows path conventions, `.exe` suffix, and proc-macro DLL extension.
2. **Fix mir-json-translate-libs Windows support** — `src/bin/mir-json-translate-libs.rs`: use `cmd /c copy` instead of POSIX `cp`, `xcopy` for directory copy, conditional `lib` prefix.
3. **Auto-deduplicate getopts rmeta on Windows in build.rs** — handles a Windows-specific build-script edge case where duplicate rmeta entries appear in the dep graph.

Touches only `src/bin/*.rs` and `build.rs`; no changes to MIR analysis or libs/. Filed as the foundational PR; a follow-up draft PR (`saw/nightly-2026-03-21`) bumps the rustc toolchain on top of this branch.

### DIFF
--- a/src/bin/wrapper.rs
+++ b/src/bin/wrapper.rs
@@ -6,6 +6,7 @@ use std::iter;
 use std::path::{Path, PathBuf};
 use std::process::{self, Command};
 use std::str;
+#[cfg(unix)]
 use std::os::unix::process::CommandExt;
 
 fn die(msg: impl Display) -> ! {
@@ -25,8 +26,10 @@ const ALREADY_SET_VAR: &str = "CRUX_MIR_ALREADY_SET_PATH";
 const LIB_PATH_VAR: &str = "LD_LIBRARY_PATH";
 #[cfg(target_os = "macos")]
 const LIB_PATH_VAR: &str = "DYLD_LIBRARY_PATH";
+#[cfg(target_os = "windows")]
+const LIB_PATH_VAR: &str = "PATH";
 
-const TOOLCHAIN: &str = "nightly-2025-09-14";
+const TOOLCHAIN: &str = "nightly-2026-03-21";
 
 fn find_lib_dir() -> PathBuf {
     let output = Command::new("rustc")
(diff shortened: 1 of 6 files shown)


RyanGlScott:
Hi @AmeliaRose802! Thank you for submitting a PR. Supporting Windows is something that we would like to do, although it has been on the backburner for a while.

I've opened https://github.com/GaloisInc/mir-json/issues/275 to describe the underlying challenges in adding Windows support, and we may want to discuss what the plan should be before implementing it. I think this PR goes some of the way towards addressing problem (1) in that issue, but not issue (2). Out of curiosity, have you tested these changes on Windows? Do they work, or are there other problems that still need addressing?

src/bin/wrapper.rs
```diff
@@ -25,8 +26,10 @@ const ALREADY_SET_VAR: &str = "CRUX_MIR_ALREADY_SET_PATH";
 const LIB_PATH_VAR: &str = "LD_LIBRARY_PATH";
 #[cfg(target_os = "macos")]
 const LIB_PATH_VAR: &str = "DYLD_LIBRARY_PATH";
+#[cfg(target_os = "windows")]
+const LIB_PATH_VAR: &str = "PATH";
 
-const TOOLCHAIN: &str = "nightly-2025-09-14";
+const TOOLCHAIN: &str = "nightly-2026-03-21";
```
RyanGlScott, line 32: I would prefer not to change the toolchain as part of this PR, as doing so requires pretty substantial changes in `mir-json` to accommodate. (See https://github.com/GaloisInc/mir-json/pull/270#issuecomment-4571292791.)

PR #214, opened 2025-12-15
implement cases for TyKind::Coroutine

spernsteiner:
Adds detailed field types to the output for `TyKind::Coroutine` and `AggregateKind::Coroutine`.

This also fixes the ABI emitted for `TyKind::CoroutineClosure` functions: they use the `RustCall` ABI, similar to `TyKind::Closure` functions, but were previously being emitted with the `Rust` ABI.

Issue #199, opened 2025-12-01
Emit proper JSON tags for `Coroutine`-related constructs


RyanGlScott:
In order to add rudimentary support for `async` functions in `crucible-mir` (see https://github.com/GaloisInc/crucible/issues/1626), we will need to ensure that coroutine-related language constructs emit unique JSON tags besides just `"unsupported"` or `"Unsupported"`. This means that the following parts of the code will need to be emit `CoroutineClosure`:

## Replacing uses of "unsupported"

The following places in the code should use the `"CoroutineClosure"` tag:

https://github.com/GaloisInc/mir-json/blob/e41330c23c06ab37e18a81efdda526cd1cccbba1/src/analyz/ty_json.rs#L609-L611

The following code should use a `"ConstructCoroutineInClosureShim"` tag:

https://github.com/GaloisInc/mir-json/blob/e41330c23c06ab37e18a81efdda526cd1cccbba1/src/analyz/ty_json.rs#L438-L440

The following code should use a `"coroutine_closure"` tag:

https://github.com/GaloisInc/mir-json/blob/e41330c23c06ab37e18a81efdda526cd1cccbba1/src/analyz/ty_json.rs#L1389-L1391

## Replacing incorrect implementations

https://github.com/GaloisInc/mir-json/blob/e41330c23c06ab37e18a81efdda526cd1cccbba1/src/analyz/to_json.rs#L264-L265

This uses the tag `"Coroutine"` for both `CoroutineClosure`s and `Coroutine`s, which I suspect is a copy-paste error. We should use the tag `"CoroutineClosure"` for the former.

spernsteiner:
I added the `TyKind::CoroutineClosure` cases in #205.

`InstanceKind::ConstructCoroutineInClosureShim` sounds complicated.  [Docs are here](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/ty/enum.InstanceKind.html#variant.ConstructCoroutineInClosureShim), but they only mention using it to implement `FnOnce::call_once`, whereas [this section in the rustc-dev-guide](https://rustc-dev-guide.rust-lang.org/coroutine-closures.html#instance-resolution) says it may also be used to implement `Fn::call`/`FnMut::call_mut` (this is the `receiver_by_ref: true` case).  I'm hoping rustc will autogenerate MIR for this so we don't need to implement it ourselves.

[...]

PR #272, opened 2026-05-26
ty_json: include shim_kind in Unsupported instance emissions

AmeliaRose802:
## Problem

Today `mir-json` emits `{"kind": "Unsupported"}` for six instance variants it does not yet model:

- `ConstructCoroutineInClosureShim`
- `ThreadLocalShim`
- `FutureDropPollShim`
- `FnPtrAddrShim`
- `AsyncDropGlueCtorShim`
- `AsyncDropGlue`

Downstream consumers (crucible-mir) cannot tell *which* shim was elided. If one of these shims is actually called during verification, the only diagnostic the user gets is "shim was Unsupported" — they can't tell whether the problem is an async drop glue, a thread-local, or something else, and each of those has a completely different workaround.

## Background

[Issue #199](https://github.com/GaloisInc/mir-json/issues/199) (filed by @RyanGlScott in Dec 2025) originally asked for distinct JSON tags on the shim variants as well as on the coroutine type constructors. That issue was marked complete after PRs #205 / #214 added proper tags for the *type-level* `Coroutine` / `CoroutineClosure` constructors, but the six `InstanceKind::*Shim` variants listed above were left emitting plain `{"kind": "Unsupported"}`. This PR finishes that work.

## Fix

Include a `shim_kind` field in each `Unsupported` JSON emission, naming the original `InstanceKind` variant. Purely additive: older crucible-mir versions that don't read `shim_kind` simply ignore it.

The tag itself has no semantic effect — it's purely for error-message quality at the point a shim is actually invoked. Without it, a user whose verified function transitively reaches one of these shims sees "Cannot call unsupported shim" with no indication of which kind, leaving them to guess between six unrelated workarounds. With it, the deferred error can name the specific shim ("Cannot call `AsyncDropGlue` shim …"), which is searchable, actionable, and reportable.

## Diff

```
src/analyz/ty_json.rs | 6 ++++++
1 file changed, 6 insertions(+)
```

## Companion PR

- [GaloisInc/crucible#1840](https://github.com/GaloisInc/crucible/pull/1840) — accepts the new `shim_kind` field on the crucible-mir side and uses it for clearer error messages if an `IkUnsupported` instance is ever called. (That PR also makes `mir_load_module` tolerant of `"Unsupported"` instead of failing the whole module load — which is what currently blocks SAW from running on any crate whose dependency graph touches async, thread-locals, or function-pointer addressing.)

Filed as draft for upstream review.

### DIFF
--- a/src/analyz/ty_json.rs
+++ b/src/analyz/ty_json.rs
@@ -455,21 +455,27 @@ impl<'tcx> ToJson<'tcx> for ty::Instance<'tcx> {
             },
             ty::InstanceKind::ConstructCoroutineInClosureShim { .. } => json!({
                 "kind": "Unsupported",
+                "shim_kind": "ConstructCoroutineInClosureShim",
             }),
             ty::InstanceKind::ThreadLocalShim(_def_id) => json!({
                 "kind": "Unsupported",
+                "shim_kind": "ThreadLocalShim",
             }),
             ty::InstanceKind::FutureDropPollShim(_def_id, _proxy_cor_ty, _impl_cor_ty) => json!({
                 "kind": "Unsupported",
+                "shim_kind": "FutureDropPollShim",
             }),
             ty::InstanceKind::FnPtrAddrShim(_def_id, _ty) => json!({
                 "kind": "Unsupported",
+                "shim_kind": "FnPtrAddrShim",
             }),
             ty::InstanceKind::AsyncDropGlueCtorShim(_def_id, _ty) => json!({
                 "kind": "Unsupported",
+                "shim_kind": "AsyncDropGlueCtorShim",
             }),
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Thanks for the PR, @AmeliaRose802!

I like the direction of travel here, but I would go about this a slightly different way. If we are going to parse these shims on the Crucible side, then instead of emitting an `"kind": "Unsupported"` tag for each shim, we should just emit the corresponding shim name (e.g., `"kind": "ConstructCoroutineInClosureShim"`). Then, on the Crucible side, we should parse each shim type into its own `InstanceKind`, and then we should gracefully handle the shims when translating `Instance`s into Crucible. That will make it clearer that Crucible actually does "support" these shims (albeit only in a trivial sense).

PR #276, opened 2026-06-12
alloc: don't assume `reallocate` happens in-place

samcowger:
This gives `crucible-mir` more flexibility in how it chooses to implement reallocation. This flexibility will be useful for
https://github.com/GaloisInc/crucible/pull/1842 in particular, whose changes make `crucible-mir` more brittle to overwriting a smaller allocation with a larger one.


spernsteiner (approved):
Looks good, but please don't merge until you have a matching crucible PR ready to do

RyanGlScott (approved):

PR #277, opened 2026-06-15
update to nightly 2026-03-21

spernsteiner:
Updates the mir-json implementation and all included libraries.

Some of the previous library patches change behavior that wasn't exercised by the crux-mir test suite.  The corresponding crucible PR (https://github.com/GaloisInc/crucible/pull/1851) adds tests for most of those.  One patches has been dropped because I couldn't come up with a test that breaks without the patch and works with it:

* c363a8c85 "Remove the dynamic CPU support detection in `memchr` package": called only from the `object` library, which is used for parsing object files in `addr2line` and `backtrace`.  I think all uses of this are unreachable for other reasons: `panic!` implementation functions are overridden in crucible-mir, and the `std::backtrace::Backtrace` type is unsupported.  The similarly-named `core::slice::memchr` module is exercised by `conc_eval/ffi/c_str_from_bytes_with_nul.rs` but doesn't use the `memchr` crate.


RyanGlScott:
I'd be surprised if you could fully drop the thread parking patch (https://github.com/GaloisInc/mir-json/commit/302330a048f5745db5444e5a19ae9c2813bf9864). I think part of the issue is that the thread parking implementation is significantly gnarlier on macOS than it is on other Unix-like OSes (e.g., Linux), so if you tested this on Linux, you wouldn't have noticed the issue. I noticed that there are macOS-only test failures on https://github.com/GaloisInc/crucible/pull/1851 that don't arise on the Linux jobs, and I would be willing to bet that the use of macOS-specific thread parking is partly to blame.

(Sorry for not checking in a test case which explained this at the time---my bad!)

Similarly, I suspect that the `memchr` patch actually is needed, although I have lost the original context for why it was required. (@glguy, do you remember?)

src/analyz/ty_json.rs
```diff
@@ -1873,7 +1873,7 @@ fn render_variant<'tcx>(
     mir: &mut MirState<'_, 'tcx>,
     adt: &AdtInst<'tcx>,
     v: &ty::VariantDef,
-    mb_discr: &Option<String>
+    discr: String
```
RyanGlScott: While you're in the neighborhood, can you leave a comment about what this `String` argument is? (Alternatively, we might consider passing a [`Discr`](https://doc.rust-lang.org/1.91.0/nightly-rustc/rustc_middle/ty/util/struct.Discr.html) value here to make this slightly less stringly-typed.)

src/analyz/ty_json.rs
```diff
@@ -624,8 +624,8 @@ impl<'tcx> ToJson<'tcx> for ty::Ty<'tcx> {
                 // TODO
                 json!({"kind": "Alias"})
             }
-            &ty::TyKind::Pat(_, _) => {
-                json!({"kind": "Unsupported"})
+            &ty::TyKind::Pat(ty, _) => {
+                json!({"kind": "Pat", "ty": ty.to_json(mir)})
```
RyanGlScott, line 632: We are intentionally not serializing the `Pattern` included in the `TyKind::Pat` because `crucible-mir` doesn't need it, correct? If so, worth adding a comment to this effect.

RyanGlScott:
After speaking with @glguy, neither of can remember what originally motivated the `memchr` patch, so I think it's fine to proceed without that patch. For what it's worth, I don't think the original motivation came about as a result of macOS/Linux differences, so that isn't a viable explanation.

src/analyz/ty_json.rs
spernsteiner: Switched to `Discr` in 61d88f38

src/analyz/ty_json.rs
spernsteiner: Added a comment in 61d88f38 about not serializing `TyKind::Pat` patterns

spernsteiner:
I reapplied 302330a048f5745db5444e5a19ae9c2813bf9864, which replaces thread park/unpark operations with no-ops.  The existing implementation worked on Linux but only in the uncontended case—in the case where the thread needs to wait, it performs an unsupported syscall.  I've updated the submodule pointer in https://github.com/GaloisInc/crucible/pull/1851, so we'll see if the macOS tests now pass.

RyanGlScott (approved):

PR #290, opened 2026-07-20
Add a command-line tool for managing 'mir-json' installations.

### LINKED ISSUE #252: UI: add scripted interface for (re)building tools/libraries
For developers of downstream tooling (e.g. SAW) who only rarely need to build or update `mir-json` and its output, the process we lay out in the README for doing so feels a little more finicky than is necessary. I would prefer for there to be a single-line command that will build a standard configuration of `mir-json` tools (`cargo install --path . --locked`), removing any old copies as needed (`cargo clean`), and to use it to generate MIR for our copy of the standard libraries (`mir-json-translate-libs`), also removing any old copies as needed (`rm -rf rlibs rlibs_real`). It would be nice if this approach could also automatically define `{CRUX,SAW}_RUST_LIBRARY_PATH`, and/or explicitly prompt a user to do so themselves, and/or warn them about any existing definitions of those variables that point elsewhere.

I imagine we could do this via `make`, or possibly by augmenting our `build.rs`.

aschwerdfeger-galois:
This pull request introduces a command-line tool, `mir-json-up`, for managing `mir-json` installations roughly along the lines of `rustup`, individual installations being identified by a triple of (1) the `mir-json` commit ref, (2) the Rust toolchain version, and (3) the target triplet.

It has been tested to the level of installing a single version of `mir-json` and successfully running the tests via `run-all.sh` on a couple of Linux distributions and macOS.

Known caveats:

* The script's method of calculating the target triplet could be improved upon.
* There is no proper mapping between `mir-json` versions and Rust toolchains, or anything preventing the user from trying to build `mir-json` on the wrong toolchain.

Fixes https://github.com/GaloisInc/mir-json/issues/252.

### DIFF
--- a/mir-json-up/mir-json-up
+++ b/mir-json-up/mir-json-up
@@ -1,0 +1,135 @@
+#!/usr/bin/env bash
+
+set -e
+
+require_tool() {
+    local TOOL_EXEC="$1"
+    local TOOL_NAME="$2"
+    set +e
+    TOOL_PATH=$(which "${TOOL_EXEC}")
+    set -e
+    if [[ "$?" != 0 ]]; then
+        echo "This script requires ${TOOL_NAME}"
+        exit 1
+    fi
+}
+
+require_rust_component() {
+    local COMPONENT_NAME="$1"
+    if [[ $(rustup component list --toolchain "${MIR_JSON_RUST_TOOLCHAIN}" --installed | grep -c "^${COMPONENT_NAME}") == 0 ]]; then
+        echo "Building mir-json requires the Rust component ${COMPONENT_NAME}; install it with"
+        echo "  rustup component add --toolchain \"${MIR_JSON_RUST_TOOLCHAIN}\" ${COMPONENT_NAME}"
+        exit 1
+    fi
+}
+
+# Need 'git' to download 'mir-json', 'cc' to run the compiler,
+# and 'rustup' to handle Rust dependencies.
+check_tools() {
+    require_tool cc "a C compiler"
+    require_tool git "Git"
+    require_tool rustup "Rustup"
+}
+
+check_rust_toolchain() {
+    if [[ $(rustup toolchain list | grep "${MIR_JSON_RUST_TOOLCHAIN}") == "" ]]; then
+        echo "Building mir-json requires the Rust toolchain ${MIR_JSON_RUST_TOOLCHAIN}; install it with"
+        echo "  rustup toolchain install ${MIR_JSON_RUST_TOOLCHAIN}"
+        exit 1
+    fi
+}
+
+get_toolchain_dir() {
+    local TOOLCHAIN="${1:-${MIR_JSON_RUST_TOOLCHAIN}}"
+    local TARGET="${2:-${TARGET_TRIPLET}}"
+    echo "${MIR_JSON_INSTALL_DIR}/toolchains/${TOOLCHAIN}/${TARGET}"
+}
+
+get_mir_json_src() {
+    local MIR_JSON_SRC_DIR="$(get_toolchain_dir)/src"
+    mkdir -p "${MIR_JSON_SRC_DIR}"
+    cd "${MIR_JSON_SRC_DIR}"
+    if ! [[ -d mir-json ]]; then
+        git clone https://github.com/GaloisInc/mir-json.git
+    fi
+}
+
+build_mir_json() {
+    local MIR_JSON_SRC_DIR="$(get_toolchain_dir)/src"
+    local MIR_JSON_BIN_DIR="$(get_toolchain_dir)/bin"
+    local MIR_JSON_RLIBS_DIR="$(get_toolchain_dir)/rlibs"
+    mkdir -p "${MIR_JSON_BIN_DIR}"
+    cd "${MIR_JSON_SRC_DIR}/mir-json"
+    cargo +"${MIR_JSON_RUST_TOOLCHAIN}" build -r --target-dir "${MIR_JSON_BIN_DIR}" --locked
+    PATH="${MIR_JSON_BIN_DIR}/release:$PATH"
+    echo $PATH
+    mir-json-translate-libs
+    if [[ $? == 0 ]]; then
+        if [[ -e "${MIR_JSON_RLIBS_DIR}" ]]; then
+            rm -rfv "${MIR_JSON_RLIBS_DIR}"
+        fi
+        cp -RL rlibs "${MIR_JSON_RLIBS_DIR}"
+    fi
+}
+
+check_all_prerequisites() {
+    check_tools
+
+    MIR_JSON_INSTALL_DIR=${MIR_JSON_INSTALL_DIR:-~/.mir-json-up}
+    MIR_JSON_RUST_TOOLCHAIN=${MIR_JSON_RUST_TOOLCHAIN:-nightly-2025-09-14}
+    TARGET_TRIPLET=${TARGET_TRIPLET:-$(cc -dumpmachine)}
+
+    check_rust_toolchain
+
+    require_rust_component rust-src
+    require_rust_component rustc-dev
+}
+
+build_env() {
+    local TOOLCHAIN_DIR=$(get_toolchain_dir)
+    cat <<EOF >> "${MIR_JSON_INSTALL_DIR}/env"
+#!/bin/sh
+case ":\${PATH}:" in
+   *:"${TOOLCHAIN_DIR}/bin/release":*)
+       ;;
+   *)
+       export PATH="${TOOLCHAIN_DIR}/bin/release:$PATH"
+esac
+export CRUX_RUST_LIBRARY_PATH="${TOOLCHAIN_DIR}/rlibs"
+EOF
+}
+
+COMMAND="$1"
+shift 1
+
+case "$COMMAND" in
+    check)
+        check_all_prerequisites
+        echo "All prerequisites in place"
+        echo "  Rust toolchain: $MIR_JSON_RUST_TOOLCHAIN"
+        echo "  Installing into: $MIR_JSON_INSTALL_DIR"
+        echo "  Target: $TARGET_TRIPLET"
+        ;;
+    build)
+        check_all_prerequisites
+        get_mir_json_src
+        build_mir_json
+        build_env
+        echo "Successfully built mir-json"
+        echo "To configure the current shell, run . ${MIR_JSON_INSTALL_DIR}/env"
+        ;;
+    run)
+        check_all_prerequisites
+        if [[ ! -e "${MIR_JSON_INSTALL_DIR}/env" ]]; then
+            echo "Cannot run before building: ${MIR_JSON_INSTALL_DIR}/env does not exist"
+            exit 1
+        fi
+        . "${MIR_JSON_INSTALL_DIR}/env"
+        $@
+        ;;
+    *)
+        echo "Invalid command: ${COMMAND}"
+        exit 1
+        ;;
+esac
+
--- a/mir-json-up/mir-json-up
+++ b/mir-json-up/mir-json-up
@@ -41,8 +41,10 @@ require_rust_component() {
     fi
 }
 
-# Need 'git' to download 'mir-json', 'cc' to run the compiler,
-# and 'rustup' to handle Rust dependencies.
+# Need 'git' to download 'mir-json', 'cc' to run the compiler
+# (the 'mir-json' build will fail for lack of a linker if a
+# C compiler is not installed), and 'rustup' to handle Rust
+# dependencies.
 check_tools() {
     require_tool cc "a C compiler"
     require_tool git "Git"
--- a/mir-json-up/mir-json-up
+++ b/mir-json-up/mir-json-up
@@ -136,6 +136,7 @@ target_triplet_from_cc() {
            ;;
         arm64-apple-darwin*)
            echo "aarch64-apple-darwin"
+           ;;
         *)
            echo "$TRIPLET"
     esac    
--- a/mir-json-up/mir-json-up
+++ b/mir-json-up/mir-json-up
@@ -221,11 +221,11 @@ check_general_prerequisites() {
        if [[ $INITIALIZE_IF_MISSING == 0 ]]; then
             echo "Checking out source repository for mir-json ${MIR_JSON_VERSION}" 1>&2
             get_mir_json_src "${MIR_JSON_VERSION}"
+            if [[ ! -d "$(get_mir_json_src_dir ${MIR_JSON_VERSION})" ]]; then
+                echo "mir-json ${MIR_JSON_VERSION} failed to check out" 1>&2
+                exit 1
+            fi
        fi
-        if [[ ! -d "$(get_mir_json_src_dir ${MIR_JSON_VERSION})" ]]; then
-            echo "mir-json ${MIR_JSON_VERSION} failed to check out" 1>&2
-            exit 1
-        fi
     else
         if [[ ! -d "$(get_mir_json_src_dir ${MIR_JSON_VERSION})" ]]; then
             echo "mir-json ${MIR_JSON_VERSION} not checked out" 1>&2
@@ -398,8 +398,11 @@ case "$COMMAND" in
             echo "${MIR_JSON_INSTALL_DIR} does not exist" 1>&2
             exit 1
         fi
+        if [[ ! -d "${MIR_JSON_INSTALL_DIR}/toolchains" || -z $(ls -A "${MIR_JSON_INSTALL_DIR}/toolchains") ]]; then
+            exit 0
+        fi
         cd "${MIR_JSON_INSTALL_DIR}/toolchains"
-        for toolchain in $(ls -1d */*/* | grep -v "[*]/[*]/[*]"); do
+        for toolchain in $(ls -1d */*/*); do
             echo -n "$toolchain"
             if [[ -f "${MIR_JSON_INSTALL_DIR}/default/id" ]]; then
                 ACTIVE_TOOLCHAIN=$(cat "${MIR_JSON_INSTALL_DIR}/default/id")
--- a/mir-json-up/mir-json-up
+++ b/mir-json-up/mir-json-up
@@ -1,14 +1,16 @@
 #!/usr/bin/env bash
 
 set -e
+IFS='
+'
 
 require_tool() {
     local TOOL_EXEC="$1"
     local TOOL_NAME="$2"
     set +e
-    TOOL_PATH=$(which "${TOOL_EXEC}")
+    local TOOL_PATH=$(which "${TOOL_EXEC}")
     if [[ "$?" != 0 ]]; then
-        echo "This script requires ${TOOL_NAME}"
+        echo "This script requires ${TOOL_NAME}" 1>&2
         exit 1
     fi
     set -e
@@ -16,10 +18,14 @@ require_tool() {
 
 require_rust_toolchain() {
     local RUST_TOOLCHAIN="${1:-${MIR_JSON_RUST_TOOLCHAIN}}"
+    # If 'rustup toolchain list' is run in any directory containing
+    # a 'rust-toolchain.toml' file, it will automatically install the
+    # toolchain named therein. To prevent this, change to a directory
+    # with no 'rust-toolchain.toml' file.
     pushd "${MIR_JSON_INSTALL_DIR}" > /dev/null
     if [[ $(rustup toolchain list | grep "${RUST_TOOLCHAIN}") == "" ]]; then
-        echo "Building mir-json requires the Rust toolchain ${RUST_TOOLCHAIN}; install it with"
-        echo "  rustup toolchain install ${RUST_TOOLCHAIN}"
+        echo "Building mir-json requires the Rust toolchain ${RUST_TOOLCHAIN}; install it with" 1>&2
+        echo "  rustup toolchain install ${RUST_TOOLCHAIN}" 1>&2
         exit 1
     fi
     popd > /dev/null
@@ -29,8 +35,8 @@ require_rust_component() {
     local COMPONENT_NAME="$1"
     require_rust_toolchain "${MIR_JSON_RUST_TOOLCHAIN}"
     if [[ $(rustup component list --toolchain "${MIR_JSON_RUST_TOOLCHAIN}" --installed | grep -c "^${COMPONENT_NAME}") == 0 ]]; then
-        echo "Building mir-json requires the Rust component ${COMPONENT_NAME}; install it with"
-        echo "  rustup component add --toolchain \"${MIR_JSON_RUST_TOOLCHAIN}\" ${COMPONENT_NAME}"
+        echo "Building mir-json requires the Rust component ${COMPONENT_NAME}; install it with" 1>&2
+        echo "  rustup component add --toolchain \"${MIR_JSON_RUST_TOOLCHAIN}\" ${COMPONENT_NAME}" 1>&2
         exit 1
     fi
 }
@@ -46,8 +52,8 @@ check_tools() {
 check_rust_toolchain() {
     require_rust_toolchain "${MIR_JSON_RUST_TOOLCHAIN}"
     if [[ $(rustup target list --installed | grep "${TARGET_TRIPLET}") == "" ]]; then
-        echo "Target ${TARGET_TRIPLET} not installed; install with"
-        echo "  rustup target add ${TARGET_TRIPLET}"
+        echo "Target ${TARGET_TRIPLET} not installed; install with" 1>&2
+        echo "  rustup target add ${TARGET_TRIPLET}" 1>&2
         exit 1
     fi
 }
@@ -57,6 +63,21 @@ get_toolchain_dir() {
     local TOOLCHAIN="${2:-${MIR_JSON_RUST_TOOLCHAIN}}"
     local TARGET="${3:-${TARGET_TRIPLET}}"
     echo "${MIR_JSON_INSTALL_DIR}/toolchains/${VERSION}/${TOOLCHAIN}/${TARGET}"
+}
+
+toolchain_is_default() {
+    local TOOLCHAIN_DIR="${1:-$(get_toolchain_dir)}"
+
+    if [[ -f "${MIR_JSON_INSTALL_DIR}/default/id" ]]; then
+        ACTIVE_TOOLCHAIN=$(cat "${MIR_JSON_INSTALL_DIR}/default/id")
+        if [[ "${ACTIVE_TOOLCHAIN}" == "${TOOLCHAIN_DIR#${MIR_JSON_INSTALL_DIR}/toolchains/}" ]]; then
+            echo -n "0"
+        else
+            echo -n "1"
+        fi
+    else
+        echo -n "1"
+    fi
 }
 
 get_mir_json_src_dir() {
@@ -72,6 +93,7 @@ get_mir_json_src() {
     if ! [[ -d mir-json ]]; then
         git clone https://github.com/GaloisInc/mir-json.git
     fi
+    cp mir-json/rust-toolchain.toml .
     cd mir-json
     set +e
     if [[ $(git show-ref refs/tags/"${MIR_JSON_VERSION}") == "" ]]; then
@@ -82,13 +104,19 @@ get_mir_json_src() {
     popd > /dev/null
 }
 
+cleanup_mir_json_src() {
+    local VERSION="${1:-${MIR_JSON_VERSION}}"
+    local MIR_JSON_SRC_DIR=$(get_mir_json_src_dir "$VERSION")
+    rm -rf "${MIR_JSON_SRC_DIR}/mir-json"
+}
+
 create_installation_dir() {
     if [[ -d "${MIR_JSON_INSTALL_DIR}/default/bin" ]]; then
           rm -rf "${MIR_JSON_INSTALL_DIR}/default/bin"
     fi
     mkdir -p "${MIR_JSON_INSTALL_DIR}/default/bin"
     pushd "${MIR_JSON_INSTALL_DIR}/default/bin" > /dev/null
-    ln -s ../../mir-json-up
+    ln -s ../../mir-json-up mir-json-up
     if [[ ! -e "${MIR_JSON_INSTALL_DIR}/mir-json-up" ]]; then
          cp "${SCRIPT_FILE}" "${MIR_JSON_INSTALL_DIR}/mir-json-up"
     fi
@@ -104,10 +132,11 @@ create_default_symlinks() {
         rm -rf "${MIR_JSON_INSTALL_DIR}/default"
     fi
     create_installation_dir
-    echo "${TOOLCHAIN_DIR}" > "${MIR_JSON_INSTALL_DIR}/default/id"
+    local TOOLCHAINS_DIR="${MIR_JSON_INSTALL_DIR}/toolchains/"
+    echo -n "${TOOLCHAIN_DIR#${TOOLCHAINS_DIR}}" > "${MIR_JSON_INSTALL_DIR}/default/id"
     pushd "${MIR_JSON_INSTALL_DIR}/default/bin" > /dev/null
-    for execFile in $(find "${MIR_JSON_BIN_DIR}" -maxdepth 1 -type f); do
-        if [[ -x "${execFile}" ]]; then
+    for execFile in "${MIR_JSON_BIN_DIR}"/*; do
+        if [[ -f "${execFile}" && -x "${execFile}" ]]; then
             ln -s "${execFile}"
         fi
     done
@@ -125,13 +154,15 @@ build_mir_json() {
     cargo +"${MIR_JSON_RUST_TOOLCHAIN}" build -r --target "${TARGET_TRIPLET}" --target-dir "${MIR_JSON_BIN_DIR}" --locked
     PATH="${MIR_JSON_BIN_DIR}/${TARGET_TRIPLET}/release:$PATH"
     mir-json-translate-libs --target "${TARGET_TRIPLET}"
-    if [[ $? == 0 ]]; then
+    if [[ $? ]]; then
         if [[ -e "${MIR_JSON_RLIBS_DIR}" ]]; then
             rm -rfv "${MIR_JSON_RLIBS_DIR}"
         fi
-        cp -RL rlibs "${MIR_JSON_RLIBS_DIR}"
+        tar cf rlibs.tar rlibs
+        pushd "$(get_toolchain_dir)" > /dev/null
+        tar xf "${MIR_JSON_SRC_DIR}/mir-json/rlibs.tar"
     fi
-    if [[ ! -d "${MIR_JSON_INSTALL_DIR}/default/bin/mir-json" ]]; then
+    if [[ ! -f "${MIR_JSON_INSTALL_DIR}/default/bin/mir-json" ]]; then
         create_default_symlinks "$(get_toolchain_dir)"
     fi
     popd > /dev/null
@@ -140,11 +171,15 @@ build_mir_json() {
 get_toolchain_from_mir_json_version() {
     local VERSION="${1:-${MIR_JSON_VERSION}}"
     local MIR_JSON_SRC_DIR=$(get_mir_json_src_dir "${VERSION}")
-    echo $(cat "${MIR_JSON_SRC_DIR}/mir-json/rust-toolchain.toml" | grep "^channel = " | cut -f 2 -d '"')
+    echo $(cat "${MIR_JSON_SRC_DIR}/rust-toolchain.toml" | grep "^channel = " | cut -f 2 -d '"')
 }
 
 get_target_triplet_from_toolchain() {
     local RUST_TOOLCHAIN="${1:-${MIR_JSON_RUST_TOOLCHAIN}}"
+    # If 'rustup toolchain list' is run in any directory containing
+    # a 'rust-toolchain.toml' file, it will automatically install the
+    # toolchain named therein. To prevent this, change to a directory
+    # with no 'rust-toolchain.toml' file.
     pushd "${MIR_JSON_INSTALL_DIR}" > /dev/null
     if [[ $(rustup toolchain list | grep "${RUST_TOOLCHAIN}") != "" ]]; then
         echo $(rustc +"${RUST_TOOLCHAIN}" --print host-tuple)
@@ -162,7 +197,7 @@ check_installation_dir() {
        fi
     fi
     if [[ ! -d "${MIR_JSON_INSTALL_DIR}" ]]; then
-       echo "Install directory ${MIR_JSON_INSTALL_DIR} does not exist"
+       echo "Install directory ${MIR_JSON_INSTALL_DIR} does not exist" 1>&2
        exit 1
     fi
 }
@@ -177,12 +212,12 @@ check_general_prerequisites() {
     MIR_JSON_VERSION="${MIR_JSON_VERSION:-master}"
     if [[ ! -d "$(get_mir_json_src_dir ${MIR_JSON_VERSION})" ]]; then
        if [[ $INITIALIZE_IF_MISSING == 0 ]]; then
-            echo "Checking out source repository for mir-json ${MIR_JSON_VERSION}"
+            echo "Checking out source repository for mir-json ${MIR_JSON_VERSION}" 1>&2
             get_mir_json_src "${MIR_JSON_VERSION}"
        fi
     fi
     if [[ ! -d "$(get_mir_json_src_dir ${MIR_JSON_VERSION})" ]]; then
-        echo "mir-json ${MIR_JSON_VERSION} not checked out"
+        echo "mir-json ${MIR_JSON_VERSION} not checked out" 1>&2
         exit 1
     fi
 }
@@ -209,7 +244,7 @@ case ":\${PATH}:" in
    *:"${MIR_JSON_INSTALL_DIR}/default/bin":*)
        ;;
    *)
-       export PATH="${MIR_JSON_INSTALL_DIR}/default/bin:$PATH"
+       export PATH="${MIR_JSON_INSTALL_DIR}/default/bin:\${PATH}"
 esac
 export CRUX_RUST_LIBRARY_PATH="${MIR_JSON_INSTALL_DIR}/default/rlibs"
 EOF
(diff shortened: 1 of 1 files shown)


mir-json-up/mir-json-up
```diff
@@ -0,0 +1,300 @@
...
+    case "$TRIPLET" in
+        x86_64-linux-gnu)
+           echo "x86_64-unknown-linux-gnu"
+           ;;
+        arm64-apple-darwin*)
+           echo "aarch64-apple-darwin"
+        *)
+           echo "$TRIPLET"
+    esac    
+}
```
RyanGlScott: This doesn't feel like a very robust way of querying the target triple that Rust uses. I think it would be better to query the Rust compiler directly, e.g.,

```
$ rustc --version --verbose | sed -n 's/host: //p'
aarch64-apple-darwin
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,300 @@
...
+    echo "   --target-triplet     Target triplet to use (default is the output of cc -dumpmachine)"
+    echo "Commands:"
+    echo "  mir-json-up check     Check if all requirements are in place for building 'mir-json'"
+    echo "  mir-json-up init      Install 'mir-json-up'"
+    echo "  mir-json-up install   Download and build a specific version of 'mir-json' for a specific target"
+    echo "  mir-json-up list      List all version/target 'mir-json' combinations installed"
+    echo "  mir-json-up remove    Remove the given 'mir-json' version/target combination"
+    echo "  mir-json-up run       Run a command within the active 'mir-json' version/target combination"
+    echo "  mir-json-up set       Set the active 'mir-json' version/target combination"
+}
```
RyanGlScott, line 245: I think should show what to pass if you want to display the help. (`mir-json-up --help`? `mir-json-up help`?)

Also, do each of these subcommands have their own help text? For instance, it may not be obvious what to pass to `mir-json-up install`, so it might be helpful to check what the expected argument(s) are by running something akin to `mir-json-up install --help`.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,300 @@
...
+build_mir_json() {
+    local MIR_JSON_SRC_DIR="$(get_toolchain_dir)/src"
+    local MIR_JSON_BIN_DIR="$(get_toolchain_dir)/bin"
+    local MIR_JSON_RLIBS_DIR="$(get_toolchain_dir)/rlibs"
+    mkdir -p "${MIR_JSON_BIN_DIR}"
+    cd "${MIR_JSON_SRC_DIR}/mir-json"
+    cargo +"${MIR_JSON_RUST_TOOLCHAIN}" build -r --target "${TARGET_TRIPLET}" --target-dir "${MIR_JSON_BIN_DIR}" --locked
+    PATH="${MIR_JSON_BIN_DIR}/${TARGET_TRIPLET}/release:$PATH"
+    echo $PATH
+    mir-json-translate-libs
```
RyanGlScott: Should this be invoked differently if you have a custom `TARGET_TRIPLET` defined?

RyanGlScott (commented):
Awesome work! I'd like to test this some more locally before I give an in-depth review, but here are some things I noticed from a quick once-over of the script.

mir-json-up/mir-json-up
spernsteiner:
```
$ rustc --print host-tuple
x86_64-unknown-linux-gnu
```

There are a bunch more `--print` options for other platform/toolchain details if you need them.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,300 @@
...
+    if [[ ! -d "${MIR_JSON_INSTALL_DIR}/default/bin/mir-json" ]]; then
+        create_default_symlinks "$(get_toolchain_dir)"
+    fi
+}
+
+get_toolchain_from_version() {
+    local VERSION="${1:-${MIR_JSON_VERSION}}"
+    case "$VERSION" in
+       *)
+           echo "nightly-2025-09-14"
```
RyanGlScott: Now that https://github.com/GaloisInc/mir-json/pull/277 has landed, it would be nice to be able to detect `nightly-2026-03-21` as well.

mir-json-up/mir-json-up
aschwerdfeger-galois: I added `help` to the list of documented commands. Currently none of the commands have any command-specific arguments, so separate help text is not needed as yet.

mir-json-up/mir-json-up
aschwerdfeger-galois: It should. I added the argument to do that.

mir-json-up/mir-json-up
RyanGlScott: > Currently none of the commands have any command-specific arguments, so separate help text is not needed as yet.

Oh, that's a bit surprising! For instance, I wouldn't expect `mir-json-up list` to take _any_ arguments, but I would expect `mir-json-up install` to take at least one argument (the `mir-json` version).

mir-json-up/mir-json-up
aschwerdfeger-galois: This is now properly detected with the switch to `rust-toolchain.toml` as the source-of-truth.

mir-json-up/mir-json-up
aschwerdfeger-galois: I added support for command-specific help as well.

The `run` command is the only one that arguably takes any arguments that are not the "global" `--mir-json-version`, `--mir-json-rust-toolchain`, and `--target-triplet`. For commands such as `install` and `set`, the "global" arguments technically cover everything, but I am fine with also supporting a positional argument for passing in the `mir-json` version.

aschwerdfeger-galois:
The script now supports running down Rust versions and target triplets using `rust-toolchain.toml` and `rustc --print host-tuple`. I tested it on macOS and Ubuntu as far as installing two versions of `mir-json` (schema versions 11 and 12) and switching between them.

mir-json-up/mir-json-up
aschwerdfeger-galois: The script now uses `rustc --print host-tuple` for obtaining the target triplet.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+            break
+            ;;
+        help)
+            shift 1
+            display_help $@
+            exit 0
+            ;;
+        *)
+            echo "Invalid command: $1"
+            exit 1
```
RyanGlScott, line 282: Perhaps it would be useful to call `display_help` after showing `Invalid command`. For instance, I note that this message displays if you run `mir-json-up --help`, and the lack of help text can be a bit of a head-scratcher if you're not aware that you need `help` instead of `--help`.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+        *)
+            echo "Usage: mir-json-up [--install-dir DIR] [--mir-json-version GIT-REF]"
+            echo "         [--mir-json-rust-toolchain RUST-TOOLCHAIN] [--target-triplet TRIPLET] command"
+            echo "Options:"
+            echo "   --install-dir        Directory holding the 'mir-json-up' installation"
+            echo "                            (default \$HOME/.mir-json-up)"
+            echo "   --mir-json-version   Git reference for the version of 'mir-json' to use/install"
+            echo "                            (default 'master')"
+            echo "   --mir-json-rust-toolchain     Rust toolchain to use (default 'nightly-2025-09-14')"
+            echo "   --target-triplet     Target triplet to use (default is the output of cc -dumpmachine)"
```
RyanGlScott, line 233: I think the `cc -dumpmachine` part is now out of date.

RyanGlScott, line 233: A minor thing, but `--mir-json-rust-toolchain`'s help text is indented differently from the other options in this help text.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+            ;;
+        *)
+            echo "Usage: mir-json-up [--install-dir DIR] [--mir-json-version GIT-REF]"
+            echo "         [--mir-json-rust-toolchain RUST-TOOLCHAIN] [--target-triplet TRIPLET] command"
+            echo "Options:"
+            echo "   --install-dir        Directory holding the 'mir-json-up' installation"
+            echo "                            (default \$HOME/.mir-json-up)"
+            echo "   --mir-json-version   Git reference for the version of 'mir-json' to use/install"
+            echo "                            (default 'master')"
+            echo "   --mir-json-rust-toolchain     Rust toolchain to use (default 'nightly-2025-09-14')"
```
RyanGlScott, line 232: Wouldn't it be more accurate to say that the default value is whatever Rust toolchain a given `--mir-json-version`'s `rust-toolchain.toml` file specifies?

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+        echo "Successfully initialized mir-json-up"
+	    echo "To configure a shell, run or insert the following into your initialization script:"
+        echo "  . ${MIR_JSON_INSTALL_DIR}/env"
+        ;;
+    install)
+        check_all_prerequisites 0
+        if [[ ! -d "${MIR_JSON_INSTALL_DIR}" ]]; then
+            echo "${MIR_JSON_INSTALL_DIR} does not exist"
+            exit 1
+        fi
```
RyanGlScott, line 311: I propose that `mir-json-up` simply create this directory if it doesn't exist. I can understand being cautious about having a tool downloading arbitrary things from the Internet, but I think it's well within the scope of the tool to create the local directories that it needs.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+}
+
+build_env() {
+    cat <<EOF >> "${MIR_JSON_INSTALL_DIR}/env"
+#!/bin/sh
+case ":\${PATH}:" in
+   *:"${MIR_JSON_INSTALL_DIR}/default/bin":*)
+       ;;
+   *)
+       export PATH="${MIR_JSON_INSTALL_DIR}/default/bin:$PATH"
```
RyanGlScott, line 212: I think writing "`$PATH`" here will literally inject your current `PATH` into the `env` file. When I tried `mir-json-up init`, it gave me the following `env` file:

```
#!/bin/sh
case ":${PATH}:" in
   *:"/Users/rscott/.mir-json-up/default/bin":*)
       ;;
   *)
       export PATH="/Users/rscott/.mir-json-up/default/bin:/Users/rscott/.opam/default/bin:/Users/rscott/.elan/bin:/Users/rscott/.cargo/bin:/Users/rscott/.cabal/bin:/Users/rscott/.ghcup/bin:/Users/rscott/.local/bin:<snip>"
esac
export CRUX_RUST_LIBRARY_PATH="/Users/rscott/.mir-json-up/default/rlibs"
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+        ;;
+    list)
+        check_general_prerequisites
+        if [[ ! -e "${MIR_JSON_INSTALL_DIR}" ]]; then
+            echo "${MIR_JSON_INSTALL_DIR} does not exist"
+            exit 1
+        fi
+        cd "${MIR_JSON_INSTALL_DIR}/toolchains"
+        find */*/* -maxdepth 0 -type d
+        ;;
```
RyanGlScott, line 323: Can `list` also show which toolchain is the currently selected one, similarly to `rustup toolchain list`?

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+    echo "${MIR_JSON_INSTALL_DIR}/src/${VERSION}"
+}
+
+get_mir_json_src() {
+    local VERSION="${1:-${MIR_JSON_VERSION}}"
+    local MIR_JSON_SRC_DIR=$(get_mir_json_src_dir "$VERSION")
+    mkdir -p "${MIR_JSON_SRC_DIR}"
+    pushd "${MIR_JSON_SRC_DIR}" > /dev/null
+    if ! [[ -d mir-json ]]; then
+        git clone https://github.com/GaloisInc/mir-json.git
```
RyanGlScott, line 73: One thing that wasn't obvious to me until I ran the script: this will save a locally-cloned checkout of the `mir-json` repo for each toolchain that you install. There are space concerns for doing this, as the `mir-json` repo is about 490 megabytes per checkout. I wonder if `mir-json-up` should remove the `src` checkout after installation has finished (or, if you want to keep the `src` directories for some reason, have a `mir-json-up gc` command to remove old `src` directories).

mir-json-up/mir-json-up
RyanGlScott: Thanks! I like the command-specific help for `run`. If we add more command-specific options in the future, we now have the machinery for doing so.

RyanGlScott (commented):
I had the chance to test this out, and it's pretty nice so far! Here are some things I noticed during my test run.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+    pushd "${MIR_JSON_INSTALL_DIR}/default/bin" > /dev/null
+    for execFile in $(find "${MIR_JSON_BIN_DIR}" -maxdepth 1 -type f); do
+        if [[ -x "${execFile}" ]]; then
+            ln -s "${execFile}"
+        fi
+    done
+    cd "${MIR_JSON_INSTALL_DIR}/default"
+    ln -s "${MIR_JSON_RLIBS_DIR}" rlibs
+    popd > /dev/null
+}
```
RyanGlScott, line 117: What happens if `TOOLCHAIN_DIR` doesn't exist? I tried a nonsensical toolchain name like `mir-json-up set blah`, and it appeared to accept that input without raising an error (it's unclear to me if it actually did anything under the hood).

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+        ;;
+    remove)
+        check_general_prerequisites
+        if [[ ! -e "${MIR_JSON_INSTALL_DIR}" ]]; then
+            echo "${MIR_JSON_INSTALL_DIR} does not exist"
+            exit 1
+        fi
+        TOOLCHAIN_DIR="$(get_toolchain_dir)"
+        rm -rf "${TOOLCHAIN_DIR}"
+        ;;
```
RyanGlScott, line 332: One thing that wasn't obvious to me until I tried the `remove` command: if you don't specify any toolchain version explicitly, then it will remove the default toolchain. This is somewhat surprising if you're not expecting it...

To make this less surprising, I wonder if we should either (1) prompt the user before deleting the default toolchain, or (2) make the toolchain version a required argument instead of an optional flag?

mir-json-up/mir-json-up
RyanGlScott, line 323: A surprising failure mode here is if you run `mir-json-up list` right after `mir-json-up init`, which prints:

```
$ ./mir-json-up list
mir-json master not checked out
```

And then returns a non-zero exit code. I would have expected something like "No toolchains currently installed" and an exit code of 0.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
+#!/usr/bin/env bash
+
+set -e
+
+require_tool() {
+    local TOOL_EXEC="$1"
+    local TOOL_NAME="$2"
+    set +e
+    TOOL_PATH=$(which "${TOOL_EXEC}")
```
sauclovian-g, line 9: `TOOL_PATH` should probably also be `local`

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
+#!/usr/bin/env bash
+
+set -e
```
sauclovian-g, line 3: Using `set -e` is not recommended. See `https://mail-index.netbsd.org/tech-userlevel/2026/01/14/msg014878.html` and `https://mail-index.netbsd.org/tech-userlevel/2026/01/14/msg014885.html` for one reason why and maintainer guidance.

(And for a script like this, I would think exiting in random places would leave the install in an undefined state and lead to trouble; better to have explicit error paths.)

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+    if [[ "$?" != 0 ]]; then
+        echo "This script requires ${TOOL_NAME}"
+        exit 1
+    fi
+    set -e
+}
+
+require_rust_toolchain() {
+    local RUST_TOOLCHAIN="${1:-${MIR_JSON_RUST_TOOLCHAIN}}"
+    pushd "${MIR_JSON_INSTALL_DIR}" > /dev/null
```
sauclovian-g, line 19: Does this `pushd`/`popd` pair actually do anything? (e.g. we need to be in that dir to run rustup or something?) If so, please comment it; otherwise probably best to remove the logic.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+    if [[ $(rustup toolchain list | grep "${RUST_TOOLCHAIN}") == "" ]]; then
+        echo "Building mir-json requires the Rust toolchain ${RUST_TOOLCHAIN}; install it with"
+        echo "  rustup toolchain install ${RUST_TOOLCHAIN}"
```
sauclovian-g, line 22:
```suggestion
        echo "Building mir-json requires the Rust toolchain ${RUST_TOOLCHAIN}; install it with" 1>&2
        echo "  rustup toolchain install ${RUST_TOOLCHAIN}" 1>&2
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+
+set -e
+
+require_tool() {
+    local TOOL_EXEC="$1"
+    local TOOL_NAME="$2"
+    set +e
+    TOOL_PATH=$(which "${TOOL_EXEC}")
+    if [[ "$?" != 0 ]]; then
+        echo "This script requires ${TOOL_NAME}"
```
sauclovian-g, line 11:
```suggestion
        echo "This script requires ${TOOL_NAME}" 1>&2
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+    fi
+    popd > /dev/null
+}
+
+require_rust_component() {
+    local COMPONENT_NAME="$1"
+    require_rust_toolchain "${MIR_JSON_RUST_TOOLCHAIN}"
+    if [[ $(rustup component list --toolchain "${MIR_JSON_RUST_TOOLCHAIN}" --installed | grep -c "^${COMPONENT_NAME}") == 0 ]]; then
+        echo "Building mir-json requires the Rust component ${COMPONENT_NAME}; install it with"
+        echo "  rustup component add --toolchain \"${MIR_JSON_RUST_TOOLCHAIN}\" ${COMPONENT_NAME}"
```
sauclovian-g, line 33:
```suggestion
        echo "Building mir-json requires the Rust component ${COMPONENT_NAME}; install it with" 1>&2
        echo "  rustup component add --toolchain \"${MIR_JSON_RUST_TOOLCHAIN}\" ${COMPONENT_NAME}" 1>&2
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+        exit 1
+    fi
+}
+
+# Need 'git' to download 'mir-json', 'cc' to run the compiler,
+# and 'rustup' to handle Rust dependencies.
+check_tools() {
+    require_tool cc "a C compiler"
```
sauclovian-g, line 41: I don't think this is actually needed any more

mir-json-up/mir-json-up
sauclovian-g, line 73: I think this, for example, should probably check explicitly for error and clean up if the clone fails.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+    local MIR_JSON_SRC_DIR=$(get_mir_json_src_dir "$VERSION")
+    mkdir -p "${MIR_JSON_SRC_DIR}"
+    pushd "${MIR_JSON_SRC_DIR}" > /dev/null
+    if ! [[ -d mir-json ]]; then
+        git clone https://github.com/GaloisInc/mir-json.git
+    fi
+    cd mir-json
+    set +e
+    if [[ $(git show-ref refs/tags/"${MIR_JSON_VERSION}") == "" ]]; then
+        git pull
```
sauclovian-g, line 78:
```suggestion
        git fetch
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+    popd > /dev/null
+}
+
+create_installation_dir() {
+    if [[ -d "${MIR_JSON_INSTALL_DIR}/default/bin" ]]; then
+          rm -rf "${MIR_JSON_INSTALL_DIR}/default/bin"
+    fi
+    mkdir -p "${MIR_JSON_INSTALL_DIR}/default/bin"
+    pushd "${MIR_JSON_INSTALL_DIR}/default/bin" > /dev/null
+    ln -s ../../mir-json-up
```
sauclovian-g, line 91: Shouldn't this have an explicit `.` as the second arg?

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+    local MIR_JSON_BIN_DIR="${TOOLCHAIN_DIR}/bin/${TARGET_TRIPLET}/release"
+    local MIR_JSON_RLIBS_DIR="${TOOLCHAIN_DIR}/rlibs"
+
+    if [[ -d "${MIR_JSON_INSTALL_DIR}/default" ]]; then
+        rm -rf "${MIR_JSON_INSTALL_DIR}/default"
+    fi
+    create_installation_dir
+    echo "${TOOLCHAIN_DIR}" > "${MIR_JSON_INSTALL_DIR}/default/id"
+    pushd "${MIR_JSON_INSTALL_DIR}/default/bin" > /dev/null
+    for execFile in $(find "${MIR_JSON_BIN_DIR}" -maxdepth 1 -type f); do
```
sauclovian-g, line 109: `find` should always have an explicit `-print`; old versions that will find the files and not print anything without are fairly rare now but you might still  bump into one.

Although, with -maxdepth 1 you might be better off just using `for f in ${MIR_JSON_BIND_DIR}/*` and filtering the type with an additional `test -f`.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+build_mir_json() {
+    local MIR_JSON_SRC_DIR="$(get_mir_json_src_dir)"
+    local MIR_JSON_BIN_DIR="$(get_toolchain_dir)/bin"
+    local MIR_JSON_RLIBS_DIR="$(get_toolchain_dir)/rlibs"
+    mkdir -p "${MIR_JSON_BIN_DIR}"
+    pushd "${MIR_JSON_SRC_DIR}/mir-json" > /dev/null
+    cargo +"${MIR_JSON_RUST_TOOLCHAIN}" build -r --target "${TARGET_TRIPLET}" --target-dir "${MIR_JSON_BIN_DIR}" --locked
+    PATH="${MIR_JSON_BIN_DIR}/${TARGET_TRIPLET}/release:$PATH"
+    mir-json-translate-libs --target "${TARGET_TRIPLET}"
+    if [[ $? == 0 ]]; then
```
sauclovian-g, line 128: This won't work until you remove the `set -e`...

(also, `==` is a pointless bashism and best avoided)

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+        if [[ -e "${MIR_JSON_RLIBS_DIR}" ]]; then
+            rm -rfv "${MIR_JSON_RLIBS_DIR}"
+        fi
+        cp -RL rlibs "${MIR_JSON_RLIBS_DIR}"
```
sauclovian-g, line 132: Tree copies should be done with `tar`, or perhaps `pax`; the behavior of `cp -r` or `cp -R` is not reliable. We had problems with this in connection with rlibs trees before...

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+
+get_toolchain_from_mir_json_version() {
+    local VERSION="${1:-${MIR_JSON_VERSION}}"
+    local MIR_JSON_SRC_DIR=$(get_mir_json_src_dir "${VERSION}")
+    echo $(cat "${MIR_JSON_SRC_DIR}/mir-json/rust-toolchain.toml" | grep "^channel = " | cut -f 2 -d '"')
+}
+
+get_target_triplet_from_toolchain() {
+    local RUST_TOOLCHAIN="${1:-${MIR_JSON_RUST_TOOLCHAIN}}"
+    pushd "${MIR_JSON_INSTALL_DIR}" > /dev/null
```
sauclovian-g, line 148: This `pushd`/`popd` pair also doesn't seem to be doing anything.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+    local INITIALIZE_IF_MISSING=${1:-1}
+
+    MIR_JSON_INSTALL_DIR="${MIR_JSON_INSTALL_DIR:-${HOME}/.mir-json-up}"
+    if [[ ! -d "${MIR_JSON_INSTALL_DIR}" ]]; then
+       if [[ $INITIALIZE_IF_MISSING == 0 ]]; then
+            create_installation_dir
+       fi
+    fi
+    if [[ ! -d "${MIR_JSON_INSTALL_DIR}" ]]; then
+       echo "Install directory ${MIR_JSON_INSTALL_DIR} does not exist"
```
sauclovian-g, line 165:
```suggestion
       echo "Install directory ${MIR_JSON_INSTALL_DIR} does not exist" 1>&2
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+
+    MIR_JSON_VERSION="${MIR_JSON_VERSION:-master}"
+    if [[ ! -d "$(get_mir_json_src_dir ${MIR_JSON_VERSION})" ]]; then
+       if [[ $INITIALIZE_IF_MISSING == 0 ]]; then
+            echo "Checking out source repository for mir-json ${MIR_JSON_VERSION}"
+            get_mir_json_src "${MIR_JSON_VERSION}"
+       fi
+    fi
+    if [[ ! -d "$(get_mir_json_src_dir ${MIR_JSON_VERSION})" ]]; then
+        echo "mir-json ${MIR_JSON_VERSION} not checked out"
```
sauclovian-g, line 185:
```suggestion
        echo "mir-json ${MIR_JSON_VERSION} failed to check out" 1>&2
```

mir-json-up/mir-json-up
sauclovian-g:
```suggestion
       export PATH="${MIR_JSON_INSTALL_DIR}/default/bin:\$PATH"
```

sauclovian-g: There's another problem with tools like this, though, which is that if they only append to PATH it grows without bound. Ideally, it should find the first `.mir-json-up/` entry and replace it (in place, not at the front), adding only if there isn't one already.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+}
+
+display_help() {
+    case "$1" in
+        run)
+            echo "Usage: mir-json-up run [--] command-line"
+            echo "   Run the given command line within the active 'mir-json' version/target combination"
+            ;;
+        *)
+            echo "Usage: mir-json-up [--install-dir DIR] [--mir-json-version GIT-REF]"
```
sauclovian-g, line 225: It would be reasonable to make the help text a here-doc

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+            echo "  mir-json-up install   Download and build a specific version of 'mir-json' for a specific target"
+            echo "  mir-json-up list      List all version/target 'mir-json' combinations installed"
+            echo "  mir-json-up remove    Remove the given 'mir-json' version/target combination"
+            echo "  mir-json-up run       Run a command within the active 'mir-json' version/target combination"
+            echo "  mir-json-up set       Set the active 'mir-json' version/target combination"
+            ;;
+    esac
+}
+
+SCRIPT_FILE=$(realpath "$0")
```
sauclovian-g, line 247: Running `realpath` is unlikely to be necessary and, if the results end up in user-facing prints, potentially confusing.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+
+while (( "$#" )); do
+    case "$1" in
+        "--install-dir")
+            MIR_JSON_INSTALL_DIR="$2"
+            shift 2
```
sauclovian-g, line 253: This will crash with "shift: can't shift that many" (or similar) if the argument is missing. Ideally, it should check that the argument is there first and print a user-readable error message if not.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+            ;;
+        "--mir-json-version")
+            MIR_JSON_VERSION="$2"
+            shift 2
```
sauclovian-g, line 257: (same for the rest of these)

(also, users tend to expect to be able to write `--mir-json-version=foo` as well as `--mir-json-version foo`)

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+            fi
+            break
+            ;;
+        help)
+            shift 1
+            display_help $@
+            exit 0
+            ;;
+        *)
+            echo "Invalid command: $1"
```
sauclovian-g, line 281:
```suggestion
            echo "Invalid command: $1" 1>&2
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+        check_installation_dir 0
+        build_env
+        echo "Successfully initialized mir-json-up"
+	    echo "To configure a shell, run or insert the following into your initialization script:"
+        echo "  . ${MIR_JSON_INSTALL_DIR}/env"
+        ;;
+    install)
+        check_all_prerequisites 0
+        if [[ ! -d "${MIR_JSON_INSTALL_DIR}" ]]; then
+            echo "${MIR_JSON_INSTALL_DIR} does not exist"
```
sauclovian-g, line 309:
```suggestion
            echo "${MIR_JSON_INSTALL_DIR} does not exist" 1>&2
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+            exit 1
+        fi
+        get_mir_json_src
+        build_mir_json
+        ;;
+    list)
+        check_general_prerequisites
+        if [[ ! -e "${MIR_JSON_INSTALL_DIR}" ]]; then
+            echo "${MIR_JSON_INSTALL_DIR} does not exist"
```
sauclovian-g, line 318:
```suggestion
            echo "${MIR_JSON_INSTALL_DIR} does not exist" 1>&2
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+            exit 1
+        fi
+        cd "${MIR_JSON_INSTALL_DIR}/toolchains"
+        find */*/* -maxdepth 0 -type d
```
sauclovian-g, line 322: `find` should have `-print`... however, I don't think this does anything that `ls -d */*/*` wouldn't.

Also, if you have no toolchains it will print `*/*/*` and you might want to filter that.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+        ;;
+    remove)
+        check_general_prerequisites
+        if [[ ! -e "${MIR_JSON_INSTALL_DIR}" ]]; then
+            echo "${MIR_JSON_INSTALL_DIR} does not exist"
```
sauclovian-g, line 327:
```suggestion
            echo "${MIR_JSON_INSTALL_DIR} does not exist" 1>&2
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+            exit 1
+        fi
+        TOOLCHAIN_DIR="$(get_toolchain_dir)"
+        rm -rf "${TOOLCHAIN_DIR}"
+        ;;
+    run)
+        check_all_prerequisites
+        if [[ ! -e "${MIR_JSON_INSTALL_DIR}/env" ]]; then
+            echo "Cannot run before building: ${MIR_JSON_INSTALL_DIR}/env does not exist"
```
sauclovian-g, line 336:
```suggestion
            echo "Cannot run before building: ${MIR_JSON_INSTALL_DIR}/env does not exist" 1>&2
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+            exit 1
+        fi
+        . "${MIR_JSON_INSTALL_DIR}/env"
+        $@
```
sauclovian-g, line 340:
```suggestion
        "$@"
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+    set)
+        check_all_prerequisites
+        create_default_symlinks "$(get_toolchain_dir)"
+        ;;
+    help)
+        display_help "$1"
+        exit 1
+        ;;
+    *)
+        echo "Invalid command: ${COMMAND}"
```
sauclovian-g, line 351:
```suggestion
        echo "Invalid command: ${COMMAND}" 1>&2
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+        exit 1
+        ;;
+esac
```
sauclovian-g, line 354: Having no newline at the end of file is asking for trouble...

sauclovian-g (commented):
Bunch of shell programming notes. Haven't had a chance to actually try it yet...

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,354 @@
...
+    fi
+}
+
+get_toolchain_dir() {
+    local VERSION="${1:-${MIR_JSON_VERSION}}"
+    local TOOLCHAIN="${2:-${MIR_JSON_RUST_TOOLCHAIN}}"
+    local TARGET="${3:-${TARGET_TRIPLET}}"
+    echo "${MIR_JSON_INSTALL_DIR}/toolchains/${VERSION}/${TOOLCHAIN}/${TARGET}"
+}
+
```
yav, line 61: It might be nice to have a way to say: please use this specific `mir-json` source directory.  I am thinking of a use case more targeted at us developers, where you are working on `mir-json` and you want to easily switch between different versions of the standard libraries.

mir-json-up/mir-json-up
aschwerdfeger-galois: Call to `display_help` added.

mir-json-up/mir-json-up
aschwerdfeger-galois: Indentation fixed with the switch to a here-doc.

mir-json-up/mir-json-up
aschwerdfeger-galois: I added the escape character to put `$PATH` literally into the `env` script. The first case in the `case` statement handles the case in which the `mir-json-up` path is already present, so there is no growth without bound.

mir-json-up/mir-json-up
aschwerdfeger-galois: Done.

mir-json-up/mir-json-up
aschwerdfeger-galois: The `set` command does not currently accept positional arguments -- it previously ignored them, hence the behavior you saw. It now raises an error if one is provided.

The prerequisites check that is run before this function is called should flag any nonexistent toolchain, but I also added an explicit check for the existence of the `TOOLCHAIN_DIR`.

mir-json-up/mir-json-up
aschwerdfeger-galois: I added a check against removing the active version.

mir-json-up/mir-json-up
aschwerdfeger-galois: Explanatory comments added.

mir-json-up/mir-json-up
aschwerdfeger-galois: Newline added.

mir-json-up/mir-json-up
aschwerdfeger-galois: When I was testing this out on Ubuntu, if the C compiler was not installed, the `mir-json` build process would fail for lack of a linker. I expanded the comment with this information.

mir-json-up/mir-json-up
aschwerdfeger-galois: Explicit second argument added.

mir-json-up/mir-json-up
aschwerdfeger-galois: I removed the `==`.

mir-json-up/mir-json-up
aschwerdfeger-galois: `tar` is now used to do the copy.

mir-json-up/mir-json-up
aschwerdfeger-galois: Switched to `ls` and added an explicit check against the no-toolchain case.

mir-json-up/mir-json-up
aschwerdfeger-galois: The reference to `cc -dumpmachine` has been removed.

mir-json-up/mir-json-up
aschwerdfeger-galois: The script now removes the `mir-json` repository when the build is completed, retaining only the `rust-toolchain.toml` file.

mir-json-up/mir-json-up
aschwerdfeger-galois: Extra checks are now performed for the case of no toolchains being installed, so this message and the nonzero exit status no longer appear.

mir-json-up/mir-json-up
aschwerdfeger-galois: I am hesitant to have any command other than `init` create the install directory, since `init` creates a copy of the script itself within the install directory, and this version is what should be used for running any other commands.

aschwerdfeger-galois:
Here are the issues with `mir-json-up` that have not yet been addressed, and should be resolved and/or split off into separate tickets if this is merged now:

* Whether to create the install directory automatically on commands other than `init`.
* Cease using `set -e` to short-circuit in case of an error (should be done, but may be a big refactoring job).
* Allow `=` separators as well as spaces in named arguments on the command line.
* Allow the use of a custom `mir-json` source directory.

mir-json-up/mir-json-up
RyanGlScott: Fair enough, I suppose checking for the presence of `MIR_JSON_INSTALL_DIR` is really a proxy for checking that that user already performed `mir-json-up init` (and everything that command entails).

RyanGlScott:
I think the first bullet point is not that big of a deal, so I am happy to go with `mir-json-up`'s current behavior. I agree that everything else can be split off into separate issues.

mir-json-up/mir-json-up
kquick: n.b. The `getopt` command can be used to support standardized option parsing of both long and short options as an alternative to the `while` loop parsing here.

mir-json-up/mir-json-up
sauclovian-g: It might make sense for these error messages to say "You need to run mir-json-up init first".

mir-json-up/mir-json-up
sauclovian-g: Hmm. Does it actually need the compiler, or just binutils? Not that many people are likely to have one without the other, so maybe it doesn't matter...

spernsteiner: `rustc` calls `cc` for the linking step (on most platforms), so that `cc` can find and pull in the appropriate libc and such.

mir-json-up/mir-json-up
```diff
@@ -1,25 +1,31 @@
...
     fi
     set -e
 }
 
 require_rust_toolchain() {
     local RUST_TOOLCHAIN="${1:-${MIR_JSON_RUST_TOOLCHAIN}}"
+    # If 'rustup toolchain list' is run in any directory containing
+    # a 'rust-toolchain.toml' file, it will automatically install the
+    # toolchain named therein. To prevent this, change to a directory
+    # with no 'rust-toolchain.toml' file.
```
sauclovian-g, line 25: picard_facepalm.jpg

mir-json-up/mir-json-up
```diff
@@ -125,13 +161,15 @@ build_mir_json() {
     cargo +"${MIR_JSON_RUST_TOOLCHAIN}" build -r --target "${TARGET_TRIPLET}" --target-dir "${MIR_JSON_BIN_DIR}" --locked
     PATH="${MIR_JSON_BIN_DIR}/${TARGET_TRIPLET}/release:$PATH"
     mir-json-translate-libs --target "${TARGET_TRIPLET}"
-    if [[ $? == 0 ]]; then
+    if [[ $? ]]; then
```
sauclovian-g: Switching from `[[ $? == 0 ]]` to `[[ $? ]]` flipped the sense of the test, and I don't think you meant that.

I'd suggest `[ $? = 0 ]`

mir-json-up/mir-json-up
```diff
@@ -125,13 +161,15 @@ build_mir_json() {
...
         if [[ -e "${MIR_JSON_RLIBS_DIR}" ]]; then
             rm -rfv "${MIR_JSON_RLIBS_DIR}"
         fi
-        cp -RL rlibs "${MIR_JSON_RLIBS_DIR}"
+        tar cf rlibs.tar rlibs
+        pushd "$(get_toolchain_dir)" > /dev/null
```
sauclovian-g, line 171: and, I think this has no matching `popd` :-(

mir-json-up/mir-json-up
aschwerdfeger-galois: `[ $? = 0 ]` adopted.

aschwerdfeger-galois:
All comments are now addressed except for the three that will be broken out into their own issues. I kicked tires on the latest version by installing two different versions of `mir-json` and running `tests/run-all.sh` with each of them in turn, and I think this is ready to merge.

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,458 @@
+#!/usr/bin/env bash
+
+set -e
+IFS='
+'
```
sauclovian-g, line 6:
```suggestion
# Remove space from IFS to avoid splitting filenames that contain spaces.
IFS='
'
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,458 @@
...
+    popd > /dev/null
+}
+
+create_default_symlinks() {
+    local TOOLCHAIN_DIR="$1"
+    local MIR_JSON_BIN_DIR="${TOOLCHAIN_DIR}/bin/${TARGET_TRIPLET}/release"
+    local MIR_JSON_RLIBS_DIR="${TOOLCHAIN_DIR}/rlibs"
+
+    if [[ ! -d "${TOOLCHAIN_DIR}" ]]; then
+        echo "Cannot set active toolchain: ${TOOLCHAIN_DIR} does not exist"
```
sauclovian-g:
```suggestion
        echo "Cannot set active toolchain: ${TOOLCHAIN_DIR} does not exist" 1>&2
```

mir-json-up/mir-json-up
```diff
@@ -0,0 +1,458 @@
...
+        if [[ -e "${MIR_JSON_RLIBS_DIR}" ]]; then
+            rm -rf "${MIR_JSON_RLIBS_DIR}"
+        fi
+        cd rlibs
+        tar cf "${MIR_JSON_SRC_DIR}/mir-json/rlibs.tar" .
+        pushd "$(get_toolchain_dir)" > /dev/null
+        mkdir rlibs
+        cd rlibs
+        tar xf "${MIR_JSON_SRC_DIR}/mir-json/rlibs.tar"
+        popd > /dev/null
```
sauclovian-g, line 175: I think this whole thing could be simplified to
```
tar cf - rlibs | (cd $(get_toolchain_dir) && tar xf -)
```
but leave it for later...

(also you might want -p in the tar invocations)

sauclovian-g (approved):
Two more little things (suggested comment, and I found another error message that goes to stdout), plus a note for later.

PR #292, opened 2026-07-20
Use `nightly-2026-03-21` consistently in docs and CI

RyanGlScott:
A follow-up to #277.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ env:
   # sometimes happen when (for example) the OS version is changed but older .so
   # files are cached, which can have various unintended effects.
   CACHE_VERSION: 1
-  RUST_TOOLCHAIN: "nightly-2025-09-14"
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ env:
   # sometimes happen when (for example) the OS version is changed but older .so
   # files are cached, which can have various unintended effects.
   CACHE_VERSION: 1
-  RUST_TOOLCHAIN: "nightly-2025-09-14"
+  RUST_TOOLCHAIN: "nightly-2026-03-21"
```
spernsteiner, line 15: I guess having the wrong nightly here doesn't cause CI to fail because the `rust-toolchain.toml` causes rustup to install the correct nightly automatically when needed.  It just makes the CI slower, since it will install the wrong nightly first before installing the right one.

spernsteiner (approved):
Oops—thanks for the fix!

PR #294, opened 2026-07-22
link: read all inputs into memory up front

spernsteiner:
There were previously several issues with the way `mir-json-dce` handled the input `.mir` files:

* The initial read of the tar (`.mir`) file tried to skip over the `crate.json` entry, but the `tar` library we use cannot seek past an entry, so the data would be read and then discarded.
* `index.cbor` was being read a few bytes at a time, with no buffering (evident when running `strace mir-json-dce foo.mir`)
* JSON chunks were being read one at a time, with buffering on the output side but not on the input.  This means one OS-level read operation for every item in every table of the final output.

This branch changes the behavior of `mir-json-dce` to read the entire `.mir` file into memory, so no further read operations are needed when reading the tar headers, parsing `index.cbor`, or copying JSON chunks to the output.  This gives about a 3x speedup when running `mir-json-dce` in isolation, and reduces the running time of the crux-mir test suite (with no cached `.mir` files) from 15 minutes to 10 minutes on my machine.

### DIFF
--- a/src/bin/mir-json-callgraph.rs
+++ b/src/bin/mir-json-callgraph.rs
@@ -12,7 +12,7 @@ extern crate mir_json;
 
 use std::collections::{HashMap, HashSet};
 use std::env;
-use std::fs::File;
+use std::fs;
 use std::io;
 use serde_json::Value as JsonValue;
 use mir_json::lib_util::{StringId, InternTable};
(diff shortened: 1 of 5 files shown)


spernsteiner:
The downside of this is it does increase the linker's memory usage by about 250MB (the sum of all `rlibs/*.mir` files).  If this is an issue, we could try `mmap`-ing the inputs instead of reading them into `Vec<u8>`.

RyanGlScott (approved):
This looks reasonable to me. I'm not sure if the extra 250 MB in memory usage is enough to make a difference in practice (what is the maximum memory use?), but if you think it is, then we could file a follow-up issue about using `mmap`.

spernsteiner:
I'm hoping the memory usage won't be an issue.  If it manages to hit 2GB on a complicated Cargo build, that should still be okay because the other big tools (crux/saw and the solver) aren't running yet at that point in time.

Issue #291, opened 2026-07-20
mir-json doesn't export intrinsic fallbacks


spernsteiner:
Some intrinsics have simplified implementations in terms of other Rust features, which could be useful for verification ([source](https://doc.rust-lang.org/1.91.1/unstable-book/language-features/intrinsics.html#intrinsics-with-fallback-logic)):

> Many intrinsics can be written in pure rust, albeit inefficiently or without supporting some features that only exist on some backends. Backends can simply not implement those intrinsics without causing any code miscompilations or failures to compile.

mir-json previously exported these, but as of #227 it seems like it no longer does (see https://github.com/GaloisInc/crucible/pull/1851#discussion_r3608267668 for details).  We should try restoring this functionality, as it might let us skip writing overrides for some intrinsics.

RyanGlScott:
> mir-json previously exported these, but as of https://github.com/GaloisInc/mir-json/issues/227 it seems like it no longer does

What is significant about #227 that caused this to change? It's not obvious to me how #227 is related to intrinsic fallbacks.

spernsteiner:
I'm assuming something changed inside rustc between the old nightly and the new one.  I'll have to investigate further to try and get a more definitive answer.

RyanGlScott:
Reading the discussion at https://github.com/GaloisInc/crucible/pull/1851#discussion_r3608267668, I'm starting to wonder if this was a quirk that was unique to the `rustc 1.91.0-nightly (02c7b1a7a 2025-09-13)` toolchain. If I try the same experiment with `rustc 1.96.0-nightly (ac7f9ec7d 2026-03-20)` (what `mir-json` uses as of the time of writing):

```rs
#[crux::test]
pub fn f() -> (u8, u8) {
    5u8.carrying_mul_add(2, 0, 0)
}
```

Then I get the following `crux-mir --print-mir` output:

```
$ cabal -v0 run exe:crux-mir -- test.rs --print-mir
FNs
fn core/a3ff0e6a::num[0]::{impl#6}[0]::carrying_mul_add[0]( _1 : u8
, _2 : u8
, _3 : u8
, _4 : u8 ) -> (u8, u8) {
   let mut _0 : (u8, u8);
   bb0: {
      call( core/a3ff0e6a::intrinsics[0]::carrying_mul_add[0]::_instb5fbeefa9be52936[0]( copy(_1)
      , copy(_2)
      , copy(_3)
      , copy(_4) )
      , (_0, bb1) )
   }
[...]

PR #295, opened 2026-07-23
emit intrinsic fallback bodies

spernsteiner:
Some intrinsics have simplified implementations in terms of other Rust features, which could be useful for verification ([source](https://doc.rust-lang.org/1.91.1/unstable-book/language-features/intrinsics.html#intrinsics-with-fallback-logic)):

> Many intrinsics can be written in pure rust, albeit inefficiently or without supporting some features that only exist on some backends. Backends can simply not implement those intrinsics without causing any code miscompilations or failures to compile.

mir-json previously exported these bodies, but after upgrading from nightly-2025-09-14 to nightly-2026-03-21, it no longer does.  The older nightly used `InstanceKind::Item` for intrinsics with bodies and `InstanceKind::Intrinsic` for those without, while the newer one uses `Intrinsic` for all intrinsics, which causes mir-json to skip emitting the function.

This branch restores output of intrinsic fallback bodies by switching from `InstanceKind::Intrinsic` to `InstanceKind::Item` when it detects that the intrinsic has a body.  This is probably not an intended use of `Instance` or `tcx.instance_mir`, but it seems to work.

### DIFF
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -1106,14 +1106,28 @@ fn emit_instance<'tcx>(
                             return Ok(());
                         }
                     }
+                    ty_inst
                 },
-                // These variants are unsupported by the `mir_shims` query, which backs
+                // Virtual shims are unsupported by the `mir_shims` query, which backs
                 // `instance_mir`.
-                ty::InstanceKind::Virtual(..) |
-                ty::InstanceKind::Intrinsic(..) => return Ok(()),
-                _ => {},
+                ty::InstanceKind::Virtual(..) => return Ok(()),
+                // Intrinsics are also unsupported by `instance_mir`, but for intrinsics that have
+                // fallback bodies, we can use a fake `InstanceKind::Item` to get it to emit the
+                // intrinsic as a normal function using the fallback body.
+                ty::InstanceKind::Intrinsic(def_id) => {
+                    let Some(intrin) = tcx.intrinsic(def_id) else { return Ok(()) };
+                    if intrin.must_be_overridden {
+                        // `must_be_overridden` means the intrinsic has no fallback body.
+                        return Ok(());
+                    }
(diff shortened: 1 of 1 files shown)


RyanGlScott:
I'm not sure I understand what effect this patch has. For instance, given the program in https://github.com/GaloisInc/mir-json/issues/291#issuecomment-5060430796, does this cause a different implementation of `carrying_mul_add` to be generated? Or does this only affect functions that lack intrinsic-based definitions on a particular configuration?

spernsteiner:
> given the program in [#291 (comment)](https://github.com/GaloisInc/mir-json/issues/291#issuecomment-5060430796), does this cause a different implementation of `carrying_mul_add` to be generated?

I haven't checked this program specifically, but it should go back to the old version from https://github.com/GaloisInc/crucible/pull/1851#discussion_r3608267668.  This shouldn't affect the crux-mir behavior for `carrying_mul_add`, since the `TransCustom` override will replace the call to `intrinsics::carrying_add_mul` before it can call into `intrinsics::fallback::carrying_add_mul`.

RyanGlScott (approved):
Ah, thanks for clarifying! I overlooked the fact that the intrinsic function itself (`core::intrinsics::carrying_mul_add`) was actually given the fallback definition. In that case, I agree that this is the change that we want.

Relatedly, do you think that we still need the `crucible-mir` override for `carrying_mul_add` now that we have a fallback implementation that `crucible-mir` can simulate?

spernsteiner:
> do you think that we still need the `crucible-mir` override for `carrying_mul_add` now that we have a fallback implementation that `crucible-mir` can simulate?

The implementations for most sizes are equivalent to what we do in the override, but for `u128` the fallback is more complex, since the Crucible implementation can easily use a `BvRepr 256`, but the Rust fallback can't.

PR #297, opened 2026-07-27
Update schema

### LINKED ISSUE #296: CI: Schema validation fails, but CI jobs succeed anyway
As seen in [this recent CI job](https://github.com/GaloisInc/mir-json/actions/runs/30053680786/job/89360880712#step:10:29), the "Validate schema" step in CI fails with lots of errors like this one:

```
+ ./simple-json-schema doc/mir-json-schema.ts --validate=./tests/issues/test0237/test.linked-mir.json
  4013--4016: JSON value shape does not match schema
                number[]
                null
    See schema:
      doc/mir-json-schema.ts:174:19--174:26
  3986--4026: JSON value shape does not match schema
                null
                {"align": 4.0, "field_offsets": null, "size": 4.0}
    See schema:
      doc/mir-json-schema.ts:140:11--140:17
  4149--4152: JSON value shape does not match schema
                number[]
                null
    See schema:
      doc/mir-json-schema.ts:174:19--174:26
  4122--4162: JSON value shape does not match schema
                null
                {"align": 8.0, "field_offsets": null, "size": 8.0}
    See schema:
      doc/mir-json-schema.ts:140:11--140:17
  5313--5316: JSON value shape does not match schema
                number[]
                null
    See schema:
      doc/mir-json-schema.ts:174:19--174:26
  5286--5326: JSON value shape does not match schema
                null
                {"align": 8.0, "field_offsets": null, "size": 8.0}
    See schema:
      doc/mir-json-schema.ts:140:11--140:17
  11--12: JSON value shape does not match schema
            7.0
            11.0
    See schema:
      doc/mir-json-schema.ts:8:12--8:12
```

Despite this, the overall CI job still passes. We should ensure that CI fails properly when there is a schema validation error, and then we should update the JSON schema accordingly so that the job passes.

yav:
We haven't been updating the machine schema, so this hopefully catches it up with what the source currently emits.

Fixes #296 by making CI use version 1.2 of `simple-json-schema` which returns an error code if schema validation fails.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
 
     - name: Validate schema
       env:
-        VER: 1.1
+        VER: 1.2
         OS_TAG: ${{ matrix.os }}
         ARCH_TAG: ${{ runner.arch }}
       run: .github/ci.sh validate_json_schema
--- a/SCHEMA_CHANGELOG.md
+++ b/SCHEMA_CHANGELOG.md
@@ -3,6 +3,16 @@ The following document describes the changes to the JSON schema that
 as a changelog for the code in the `mir-json` tools themselves, which are
 versioned separately.)
 
+## 12
+
+  * Instance: `CloneShim.callees` is now emitted as `Ty?[]`.
+    Elements are `null` when instance resolution yielded no callee.
+  * Variant: the `discr` field has been **removed** from variant output.
+  * Operand: New variant `RuntimeChecks`.
+  * InlineType: New variant `Pat`.
+  * CastKind: New nullary variant `Subtype`.
+  * `PointerCoercion.ClosureFnPointer` now has a `safety` field
+
 ## 11
 
 Add `size` and `align` fields to `dyn Trait` vtables.
--- a/SCHEMA_CHANGELOG.md
+++ b/SCHEMA_CHANGELOG.md
@@ -5,8 +5,7 @@ versioned separately.)
 
 ## 12
 
-  * Instance: `CloneShim.callees` is now emitted as `Ty?[]`.
-    Elements are `null` when instance resolution yielded no callee.
+  * Instance: `ClosureShim` has been renamed to `CloneShim`.
   * Variant: the `discr` field has been **removed** from variant output.
   * Operand: New variant `RuntimeChecks`.
   * InlineType: New variant `Pat`.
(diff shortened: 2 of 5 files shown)


SCHEMA_CHANGELOG.md
```diff
@@ -3,6 +3,16 @@ The following document describes the changes to the JSON schema that
 as a changelog for the code in the `mir-json` tools themselves, which are
 versioned separately.)
 
+## 12
+
+  * Instance: `CloneShim.callees` is now emitted as `Ty?[]`.
+    Elements are `null` when instance resolution yielded no callee.
```
spernsteiner: Can this ever actually be `null`?  From [the docs on `Instance::try_resolve`](https://doc.rust-lang.org/1.96.0/nightly-rustc/rustc_middle/ty/struct.Instance.html#method.try_resolve), it sounds like the way we use it should never return `None` unless there's a compile error.  And I don't think Crucible supports this either—if a callee is missing, it won't be able to build the clone shim.

.github/workflows/ci.yml
```diff
@@ -69,7 +69,7 @@ jobs:
 
     - name: Validate schema
       env:
-        VER: 1.1
+        VER: 1.2
```
spernsteiner, line 72: Does this fix #296?

yav: Yes, I updated the schema checker to exit with a non-zero code if schema validation fails (previously it would print the errors but return 0)

SCHEMA_CHANGELOG.md
yav: It's just based on the type (https://github.com/GaloisInc/mir-json/blob/a22e8840d263097ea4650cf090e6756710503181/src/analyz/ty_json.rs#L440) since `inst` is an `Option` we end up with the possibility of `null`.

We could change that code to throw an error or something if we encounter `None`, then the schema could be just `Ty[]`?

spernsteiner: That's probably the thing to do, for clarity.  crux-mir will fail to parse the JSON if any of those are null: https://github.com/GaloisInc/crucible/blob/0f5a63a6fd53bb7711eeb261cbe804e6efce7376/crucible-mir/src/Mir/Mir.hs#L229

.github/workflows/ci.yml
RyanGlScott: In that case, could you please add "fixes #296" to the relevant commit's commit message?

RyanGlScott (approved):

PR #299, opened 2026-08-03
Track crate SVH hashes in MIR

Ra1issa:
This MR store the crate and its dependencies SVH hashes in MIR files in order to keep track of what crates have been updated later down the line. We will specifically be using this to improve the compile time of some CI builds in `crux-mir` that don't need to be recompiled if neither the `mir-json` nor the crates used have changed.

### DIFF
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -1369,6 +1369,70 @@ pub struct AnalysisData<O> {
     pub output: O,
 }
 
+/*
+Note [CrateDepKind::MacrosOnly]
+~~~~~~~~~~~~~~~~~~~~~
+In what follows, we store the crate dependencies of the current crate for later linker 
+related validation and exclude crates flagged as CrateDepKind::MacrosOnly.
+The CrateDepKind enum has some misleading documentation about what is considered a 
+CrateDepKind::MacrosOnly crate. The documentation says:
+
+    "A dependency that is only used for its macros."
+
+This is actually slightly inaccurate. CrateDepKind::MacrosOnly is dedicated to proc_macro crates,
+rather than other crates where the user only imports macros, those being flagged as CrateDepKind::Unconditional. 
+
+In order to test this, we created a crate called hello_world that looked like this:
+
+```
+#[macro_export]
+macro_rules! foo {
+    ($msg:expr) => {
+        $msg
+    };
+}
+
+pub fn add_one(x: usize) -> usize{
+    return x + 1;
+}
+```
+
+Then I added the following to crux-mir's example-1:
+
+```
+use hello_world::foo;
+
+/// Find-first-set (fast implementation)
+pub fn ffs_fast(mut i: u32) -> u32 {
+    foo!("hello");
+    let mut n = 1;
+    if (i & 0xffff) == 0 { n += 16; i >>= 16; }
+    if (i & 0x00ff) == 0 { n +=  8; i >>=  8; }
+    if (i & 0x000f) == 0 { n +=  4; i >>=  4; }
+    if (i & 0x0003) == 0 { n +=  2; i >>=  2; }
+    if i != 0 {
+        return n + ((i.wrapping_add(1)) & 0x01);
+    } else {
+        return 0;
+    }
+}
+
+/// Find-first-set (reference implementation)
+pub fn ffs_ref(word: u32) -> u32 {
+    foo!("hello");
+    for i in 0 .. 32 {
+        if word & (1 << i) != 0 {
+            return i + 1;
+        }
+    }
+    return 0;
+}
+```
+The hello_world crate was flagged as CrateDepKind::Unconditional by the function below.
+For more information on why we exclude proc_macro crates to begin with, please checkou the
+note over lib_util::check_dependencies.
+*/
+
 /// Analyze the crate currently being compiled.  Returns `Ok(Some(data))` upon successfully writing
 /// the crate MIR, returns `Ok(None)` when there is no need to write out MIR (namely, when `comp`
 /// is not producing an `Exe` output), and returns `Err(e)` on I/O or serialization errors.
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -12,7 +12,7 @@ use rustc_session::config::{OutputType, OutFileName};
 use rustc_span::Span;
 use rustc_span::symbol::{Symbol, Ident};
 use rustc_abi::{self, ExternAbi};
-use std::collections::HashMap;
+use std::collections::{HashSet};
 use std::fmt::Write as FmtWrite;
 use std::io;
 use std::iter;
@@ -1367,7 +1367,7 @@ pub struct AnalysisData<O> {
     pub mir_path: PathBuf,
     pub extern_mir_paths: Vec<PathBuf>,
     pub output: O,
-    pub svh_hashes: Option<HashMap<String, String>>
+    pub svh_hashes: Option<HashSet<(String, String)>>
 }
 
 /// Analyze the crate currently being compiled.  Returns `Ok(Some(data))` upon successfully writing
@@ -1398,11 +1398,11 @@ fn analyze_inner<'tcx, O: JsonOutput, F: FnOnce(&Path) -> io::Result<O>>(
     };
     let mut out = mk_output(&mir_path)?;
 
-    let mut svh_hashes = HashMap::new();
+    let mut svh_hashes = HashSet::new();
     for &cnum in tcx.crates(()) {
-        svh_hashes.insert(
+        svh_hashes.insert((
             tcx.crate_name(cnum).to_string(),
-            tcx.crate_hash(cnum).to_string() 
+            tcx.crate_hash(cnum).to_string() )
         );
 
         let src = tcx.used_crate_source(cnum);
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -1403,10 +1403,13 @@ fn analyze_inner<'tcx, O: JsonOutput, F: FnOnce(&Path) -> io::Result<O>>(
 
     let mut dep_hashes = HashSet::new();
     for &cnum in tcx.crates(()) {
-        dep_hashes.insert(SvhHash::new(
+
+        // Exclude proc_macro crates
+        if ! tcx.crate_dep_kind(cnum).macros_only(){        
+            dep_hashes.insert(SvhHash::new(
             tcx.crate_name(cnum).to_string(),
-            tcx.crate_hash(cnum).to_string() )
-        );
+            tcx.crate_hash(cnum).to_string())
+        );}
 
         let src = tcx.used_crate_source(cnum);
         let it = src.dylib.iter()
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -12,6 +12,7 @@ use rustc_session::config::{OutputType, OutFileName};
 use rustc_span::Span;
 use rustc_span::symbol::{Symbol, Ident};
 use rustc_abi::{self, ExternAbi};
+use std::collections::HashMap;
 use std::fmt::Write as FmtWrite;
 use std::io;
 use std::iter;
@@ -1366,6 +1367,7 @@ pub struct AnalysisData<O> {
     pub mir_path: PathBuf,
     pub extern_mir_paths: Vec<PathBuf>,
     pub output: O,
+    pub svh_hashes: Option<HashMap<String, String>>
 }
 
 /// Analyze the crate currently being compiled.  Returns `Ok(Some(data))` upon successfully writing
@@ -1396,7 +1398,13 @@ fn analyze_inner<'tcx, O: JsonOutput, F: FnOnce(&Path) -> io::Result<O>>(
     };
     let mut out = mk_output(&mir_path)?;
 
+    let mut svh_hashes = HashMap::new();
     for &cnum in tcx.crates(()) {
+        svh_hashes.insert(
+            tcx.crate_name(cnum).to_string(),
+            tcx.crate_hash(cnum).to_string() 
+        );
+
         let src = tcx.used_crate_source(cnum);
         let it = src.dylib.iter()
             .chain(src.rlib.iter())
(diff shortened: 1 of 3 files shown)


spernsteiner:
There are a bunch of unnecessary rustfmt changes mixed in with the actual SVH changes for this PR.  Please undo all of the rustfmt changes so this branch contains only the SVH stuff.

src/analyz/mod.rs
```diff
@@ -1396,7 +1398,13 @@ fn analyze_inner<'tcx, O: JsonOutput, F: FnOnce(&Path) -> io::Result<O>>(
     };
     let mut out = mk_output(&mir_path)?;
 
+    let mut svh_hashes = HashMap::new();
     for &cnum in tcx.crates(()) {
+        svh_hashes.insert(
+            tcx.crate_name(cnum).to_string(),
```
spernsteiner: Are all `crate_name`s guaranteed to be distinct, or could there be collisions here?  I'm particularly wondering about these two cases:
1. The current crate depends on `foo` v1, and also depends on `foo` v2, with the latter being renamed/aliased to `foo2`.  Does `crate_name` for v2 report the alias `foo2`, or do both report the original name `foo`?
2. The current crate depends on `a` which depends on `foo` v1, and the current crate also depends on `b` which depends on `foo` v2 (no alias is needed in this case).  Do the transitive dependencies show up in `tcx.crates()` output?  And if so, do they both have a `crate_name` of `foo`?

src/lib_util.rs
```diff
@@ -48,6 +49,9 @@ pub struct CrateIndex {
 
     /// The schema version in use. (See also `SCHEMA_VER`.)
     pub version: u64,
+
+    /// Mapping from the names of crates the current crate depends on to their Svh hashes as Strings.
+    pub svh_hashes: HashMap<String, String>
```
spernsteiner: For doing checks at link time, I think we also need to record the name and hash of the current crate.  Then if any crate depends on a (name, hash) pair that isn't the "current crate" of any input file, that's an error.

src/lib_util.rs
```diff
@@ -555,10 +559,10 @@ impl<W: Write> StreamingEmitter<W> {
         Ok(se)
     }
 
-    pub fn finish(mut self) -> io::Result<(W, CrateIndex)> {
+    pub fn finish(mut self, svh_hashes: HashMap<String, String>) -> io::Result<(W, CrateIndex)> {
```
spernsteiner: Seems unfortunate to pass around `svh_hashes` everywhere like this, but I don't see a better place to put it.

Might be worth wrapping this in a struct so we can bundle up all the extra things we want to include in the crate index (the dependencies and the current crate name/hash, plus anything we may want to add later) and pass them around as a single argument.

Ra1issa:
Documenting this in case we come back to this PR in the future:

The [CrateDepKind](https://doc.rust-lang.org/stable/nightly-rustc/rustc_session/cstore/enum.CrateDepKind.html) enum has some weird documentation about what is considered a `MacrosOnly` crate. The documentation says:

> _"A dependency that is only used for its macros."_

This is actually slightly inaccurate. `MacrosOnly` is dedicated to `proc_macro` crates, rather than other crates where the user only imports `macros`. This is why [this](https://github.com/GaloisInc/mir-json/pull/299/changes/5c0fd74a3486d30cc17fc5703e219abfeeb94443) commit was done the way it was. 

To test this, I crated a crate called `hello_world` that has the following
```rust
#[macro_export]
macro_rules! foo {
    ($msg:expr) => {
        $msg
    };
}

pub fn add_one(x: usize) -> usize{
    return x + 1;
}
```
Then I added the following to my `crux-mir` [example](https://github.com/GaloisInc/crucible/blob/master/crux-mir/examples/example-1/lib.rs):
```rust
use hello_world::foo;

/// Find-first-set (fast implementation)
pub fn ffs_fast(mut i: u32) -> u32 {
    foo!("hello");
    let mut n = 1;
    if (i & 0xffff) == 0 { n += 16; i >>= 16; }
    if (i & 0x00ff) == 0 { n +=  8; i >>=  8; }
    if (i & 0x000f) == 0 { n +=  4; i >>=  4; }
    if (i & 0x0003) == 0 { n +=  2; i >>=  2; }
    if i != 0 {
        return n + ((i.wrapping_add(1)) & 0x01);
    } else {
        return 0;
    }
}

/// Find-first-set (reference implementation)
pub fn ffs_ref(word: u32) -> u32 {
    foo!("hello");
    for i in 0 .. 32 {
        if word & (1 << i) != 0 {
            return i + 1;
        }
    }
    return 0;
}
```
The `hello_world` crate was flagged as `Unconditional`.

src/lib_util.rs
```diff
@@ -627,3 +653,39 @@ pub fn finish_streaming(ms: MirStream) -> serde_cbor::Result<()> {
...
+
+impl std::error::Error for DependencyError {}
+
+/// Function that checks if the crate dependencies were provided as inputs to the linker
+pub fn check_dependencies(crate_indices: &[CrateIndex]) -> Result<(), DependencyError>{
+    let mut linker_inputs: HashSet<SvhHash>  = HashSet::new();
+    for index in crate_indices{
+        linker_inputs.insert(index.root_hash.clone());
+    }
+    for index in crate_indices{
```
RyanGlScott: Minor formatting nit picks:


```suggestion
pub fn check_dependencies(crate_indices: &[CrateIndex]) -> Result<(), DependencyError> {
    let mut linker_inputs: HashSet<SvhHash>  = HashSet::new();
    for index in crate_indices {
        linker_inputs.insert(index.root_hash.clone());
    }
    for index in crate_indices {
```

And similarly in other parts of this PR.

src/analyz/mod.rs
```diff
@@ -1366,6 +1367,8 @@ pub struct AnalysisData<O> {
     pub mir_path: PathBuf,
     pub extern_mir_paths: Vec<PathBuf>,
     pub output: O,
+    pub root_hash: Option<SvhHash>,
+    pub dep_hashes: Option<HashSet<SvhHash>>
```
RyanGlScott: Can you document under what circumstances these fields will be `None`?

src/analyz/mod.rs
```diff
@@ -1395,8 +1398,19 @@ fn analyze_inner<'tcx, O: JsonOutput, F: FnOnce(&Path) -> io::Result<O>>(
         },
     };
     let mut out = mk_output(&mir_path)?;
+    let root_hash = SvhHash::new(tcx.crate_name(LOCAL_CRATE).to_string(),tcx.crate_hash(LOCAL_CRATE).to_string());
```
RyanGlScott: This pattern of creating an `SvhHash` from a `CrateNum` is used in multiple places (here, and below in the `for` loop). It might be worth defining an `SvhHash::from_crate_num` function to avoid some duplication.

src/lib_util.rs
```diff
@@ -32,6 +35,24 @@ use tar;
 use crate::schema_ver::SCHEMA_VER;
 use crate::tar_stream::{TarStream, TarEntryStream};
 
+#[derive(Debug, Default, Serialize, Deserialize, Eq, PartialEq, PartialOrd, Ord, Hash, Clone)]
+pub struct SvhHash{
```
RyanGlScott: I had to look this up: "`Svh`" stands for "strict version hash". Could you document this?

src/lib_util.rs
```diff
@@ -48,6 +69,11 @@ pub struct CrateIndex {
 
     /// The schema version in use. (See also `SCHEMA_VER`.)
     pub version: u64,
+    /// The current crate's [`SvhHash`]
+    pub root_hash: SvhHash,
+    /// Mapping from the names of crates the current crate depends on to their Svh hashes as Strings.
+    /// The [`SvhHash`] struct consist of a tuple of a crate name and its hash
```
RyanGlScott: I'm not sure if this sentence in the documentation adds much, as it's essentially just repeating information that's readily apparent if you jump to the definition of `SvhHash`.

src/lib_util.rs
```diff
@@ -48,6 +69,11 @@ pub struct CrateIndex {
 
     /// The schema version in use. (See also `SCHEMA_VER`.)
     pub version: u64,
+    /// The current crate's [`SvhHash`]
+    pub root_hash: SvhHash,
```
RyanGlScott: If I understand correctly, this field is included in the `CrateIndex` for two reasons:

1. Extra validation that the expected dependencies were actually provided to the linker (see `check_dependencies`).
2. A way for `mir-json` users (e.g., `crux-mir`'s CI) to check if the code in a crate has changed since the last time `mir-json` was rebuilt.

Have I captured that correctly? If so, could you document this?

RyanGlScott (commented):
Thanks for working on this, @Ra1issa!

RyanGlScott:
The observation about `MacrosOnly` is a good one, and it may be worth including a comment near the affected code (linking back to your comment in https://github.com/GaloisInc/mir-json/pull/299#issuecomment-5256742177) to motivate why the code is the way it is.

src/lib_util.rs
spernsteiner: Naming this `SvhHash` seems a little misleading, since it contains both the SVH and the crate name.  Maybe it should be called something like `UniqueCrateId`?

src/lib_util.rs
spernsteiner: > Extra validation that the expected dependencies were actually provided to the linker (see `check_dependencies`).

This is required to do any validation at all.  The validation step checks whether any crate depends on something that's not included in the link (such as an outdated version of `std`), and the `root_hash` is needed to determine what's included in the link.

src/lib_util.rs
```diff
@@ -354,20 +380,20 @@ impl<W: Write> Emitter<W> {
         Ok(())
     }
 
-    pub fn finish(self) -> CrateIndex {
-        self.state.finish()
+    pub fn finish(self, dep_hashes: HashSet<SvhHash>, root_hash: SvhHash) -> CrateIndex {
```
spernsteiner: Please bundle up these two parameters into a struct (maybe call it `IndexExtra` since it's extra data to add when emitting the index) so that we only need to pass around one thing to all these various functions instead of passing two separate things.

src/analyz/mod.rs
```diff
@@ -1395,8 +1398,19 @@ fn analyze_inner<'tcx, O: JsonOutput, F: FnOnce(&Path) -> io::Result<O>>(
...
+
+    let mut dep_hashes = HashSet::new();
     for &cnum in tcx.crates(()) {
+
+        // Exclude proc_macro crates
+        if ! tcx.crate_dep_kind(cnum).macros_only(){        
+            dep_hashes.insert(SvhHash::new(
+            tcx.crate_name(cnum).to_string(),
+            tcx.crate_hash(cnum).to_string())
+        );}
```
spernsteiner: Misleading indentation

Ra1issa:
> The observation about `MacrosOnly` is a good one, and it may be worth including a comment near the affected code (linking back to your comment in [#299 (comment)](https://github.com/GaloisInc/mir-json/pull/299#issuecomment-5256742177)) to motivate why the code is the way it is.

@RyanGlScott Please checkout the last 4 commits: I have added notes and docs to capture all the things that we discovered in this PR :)

src/analyz/mod.rs
```diff
@@ -1395,8 +1460,16 @@ fn analyze_inner<'tcx, O: JsonOutput, F: FnOnce(&Path) -> io::Result<O>>(
         },
     };
     let mut out = mk_output(&mir_path)?;
+    let root_hash = UniqueCrateId::new(tcx,LOCAL_CRATE);
+
 
+    let mut dep_hashes = HashSet::new();
     for &cnum in tcx.crates(()) {
+        // Exclude proc_macro crates from the hashset of dependencies.
+        // See note above for more info.
```
RyanGlScott, line 1469: I'd prefer to mention the Note by name, as this will make the comments here more resistant to moving code around:


```suggestion
        // See Note [CrateDepKind::MacrosOnly].
```

src/analyz/mod.rs
```diff
@@ -1368,19 +1369,83 @@ pub struct AnalysisData<O> {
...
+        if word & (1 << i) != 0 {
+            return i + 1;
+        }
+    }
+    return 0;
+}
+```
+The hello_world crate was flagged as CrateDepKind::Unconditional by the function below.
+For more information on why we exclude proc_macro crates to begin with, please checkou the
+note over lib_util::check_dependencies.
```
RyanGlScott, line 1433:
```suggestion
For more information on why we exclude proc_macro crates to begin with, please check
Note [Excluding proc macro crates].
```

src/analyz/mod.rs
```diff
@@ -1395,8 +1460,16 @@ fn analyze_inner<'tcx, O: JsonOutput, F: FnOnce(&Path) -> io::Result<O>>(
...
     let mut out = mk_output(&mir_path)?;
+    let root_hash = UniqueCrateId::new(tcx,LOCAL_CRATE);
+
 
+    let mut dep_hashes = HashSet::new();
     for &cnum in tcx.crates(()) {
+        // Exclude proc_macro crates from the hashset of dependencies.
+        // See note above for more info.
+        if ! tcx.crate_dep_kind(cnum).macros_only(){        
+            dep_hashes.insert(UniqueCrateId::new( tcx, cnum));}
```
RyanGlScott, line 1471: There's some strange formatting in this code. I'd prefer something like:


```suggestion
        if !tcx.crate_dep_kind(cnum).macros_only() {        
            dep_hashes.insert(UniqueCrateId::new(tcx, cnum));
        }
```

RyanGlScott (commented):
Looks reasonable---minor comments only.

Are you ready to land this now, or do you want to test this in Crucible first before merging it?

src/analyz/mod.rs
```diff
@@ -1450,20 +1524,22 @@ fn analyze_inner<'tcx, O: JsonOutput, F: FnOnce(&Path) -> io::Result<O>>(
...
 ) -> Result<Option<AnalysisData<()>>, serde_cbor::Error> {
-    let opt_ad = analyze_inner(tcx, export_style, |_| { Ok(lib_util::Output::default()) })?;
-    let AnalysisData { mir_path, extern_mir_paths, output: out } = match opt_ad {
+    let (opt_ad, extra_data_option) = analyze_inner(tcx, export_style, |_| { Ok(lib_util::Output::default()) })?;
+    let AnalysisData { mir_path, extern_mir_paths, output: out} = match opt_ad {
         Some(x) => x,
         None => return Ok(None),
     };
 
+    let extra_data= extra_data_option.ok_or(EmptyCrateDependenciesError)?;
```
spernsteiner: If you switch the return type of `analyze_inner` from `Result<(Option<A>, Option<B>), E>` to `Result<Option<(A, B)>, E>`, then you'll always have an `extra_data` once you get past the `match opt_ad`, and can get rid of the new `EmptyCrateDependencesError` type.

Ra1issa: good catch. I tried that initially and was getting some silly error. I probably did something silly.

Ra1issa: addressed [here](https://github.com/GaloisInc/mir-json/pull/299/changes/66ea2ccae32fca01251f3c83219383054d87e7ff).

Issue #177, opened 2025-10-14
Always emit rendered constants for static items


RyanGlScott:
Currently, `mir-json` has multiple conventions regarding what JSON code to emit for static items:

1. For promoted constants (e.g., `&42`), the emitted JSON contains a `"rendered"` field that contains the rendered value of the constant.
2. For `static` declarations (e.g., `static X: i32 = 42`), the emitted JSON does _not_ contain a `"rendered"` field. Instead, the _body_ of the static item (i.e., the part to the right of the equals sign) is emitted as a separate function instance, which is referred to as the static's _initializer_. `crucible-mir` is responsible for evaluating each static initializer to a constant during the start of simulation and then populating the value of each static with the corresponding constant.

This patchwork approach reflects the fact that the rendered constant machinery is a relatively recent innovation in `rustc`, and `mir-json` has gradually been migrating some (but not all) parts of the code to use it. This issue proposes to always use the rendered constant machinery for all static items, regardless of whether they are promoted constants or `static` declarations. There are a number of advantages to doing so:

1. We would be able to treat all static items uniformly in `crucible-mir`, which would simplify some code downstream.
2. `crucible-mir`'s treatment of static initializers is somewhat buggy, as seen in https://github.com/GaloisInc/crucible/issues/1108. If we switched to using rendered constants instead of static initializers, then these sorts of issues would be fixed automatically.
3. In https://github.com/GaloisInc/saw-script/issues/2703, we want to be able to match static items against their initial values, and having the rendered constants available would greatly simplify the implementation details.

If we do this, we should be careful to ensure that we continue to handle nested static items correctly (see https://github.com/GaloisInc/mir-json/issues/129). As noted in https://github.com/GaloisInc/mir-json/issues/129#issuecomment-3387602092, we would need to adjust `try_render_ref_opty` to treat `GlobalAlloc::Static` on a nested static similar to the way it currently treats `GlobalAlloc::Memory`.

RyanGlScott:
In order to make this work, I think we will need to figure some way to leverage the `rustc` API to evaluate the `Body` of a static initializer to an `MPlaceTy`, as our rendered constant machinery requires an `MPlaceTy` as an argument:

https://github.com/GaloisInc/mir-json/blob/03bcbb1d07be8489e5148fe5eb58da8f9452a100/src/analyz/ty_json.rs#L1483-L1489

[...]

PR #302, opened 2026-08-06
Simplify away some unused code related to constant evaluation

RyanGlScott:
This PR contains a series of commits that removes some unused or redundant code in the constant evaluation machinery:

* `make_allocation_body` and `try_render_ref_opty` take `Ty` arguments, but these are redundant because these types are always equal to the ones contained in he coresponding `MPlaceTy` arguments. These `Ty` arguments are now removed.
* Remove the `emit_promoted` function. Nowadays, promoteds are only temporarily created during constant evaluation, and they never survive all the way through to the final, evaluated result. As such, there is no point in generating functions or static items for promoted things, as they will simply be eliminated later by `mir-json`'s dead code elimination.
* After `emit_promoted`'s removal, the `emit_static_decl` function now only has a single call site in `emit_static`, and the
implementation of `emit_static_decl` is not that complex. As such, we now inline `emit_static_decl` into `emit_static` to simplify things.

Progress towards #177.

### DIFF
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -947,10 +947,6 @@ fn emit_static(ms: &mut MirState, out: &mut impl JsonOutput, def_id: DefId) -> i
     emit_fn(ms, out, &name, None, mir)?;
     emit_static_decl(ms, out, &name, mir.return_ty(), tcx.is_mutable_static(def_id))?;
 
-    for (idx, mir) in tcx.promoted_mir(def_id).iter_enumerated() {
-        emit_promoted(ms, out, &name, idx, mir)?;
-    }
-
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


samcowger (approved):
Looks reasonable enough. Have you run `crux-mir{,-comp}` tests locally that rely on these changes?

RyanGlScott:
No, not yet. I'll hold off on merging this until I've had a chance to do so. (This will require first updating `crux-mir-comp` to support schema version 12.)

yav (approved):
We may want to wait until I add my PR as well, before syncing up with crucible to avoid having to do it a bunch of times.

RyanGlScott:
I tested each commit in this PR against the `crux-mir` and `crux-mir-comp` test suites. The good news is that these tests continue to pass with https://github.com/GaloisInc/mir-json/pull/302/commits/fb49502782f4c8246d58dfe81e01cb8ea4a19a45 / https://github.com/GaloisInc/mir-json/pull/302/commits/c978e3ce0347056c3ea7552938bc3563865e35dd applied. The bad news is that the `crux-mir` test suite fails with https://github.com/GaloisInc/mir-json/pull/302/commits/fec37b1d2ae333dcbe66384fe6d6a99dd7ac220f applied. Here is an example of a program which `crux-mir` crashes on when using that `mir-json` commit:

```rs
#[crux::test]
pub fn f() -> u16 {
    let v: Vec<u16> = vec![1, 2];
    v[0]
}
```
```
$ cabal -v0 run exe:crux-mir -- test.rs
test test/edf2d66e::f[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/edf2d66e::f[0] counterexamples ----
[Crux] Found counterexample for verification goal
[Crux]   ./libs/core/src/panic/location.rs:182:41: 182:51: error: in core/a3ff0e6a::panic[0]::location[0]::{impl#6}[0]::caller[0]::DUMMY_LOCATION[0]
[Crux]   Translation error in core/a3ff0e6a::panic[0]::location[0]::{impl#6}[0]::caller[0]::DUMMY_LOCATION[0]: ill-typed assignment of IntrinsicRepr MirReference [] (TyRef TyStr Immut) to StructRepr [IntrinsicRepr MirReference [], BVRepr 64] (TyRef TyStr Immut) LBase (Var {_varname = "_3", _varmut = Immut, _varty = TyRef TyStr Immut, _varIsZST = False})
[Crux]   Context:
[Crux]     ./libs/core/src/panic/location.rs:182:41: 182:51: core/a3ff0e6a::panic[0]::location[0]::{impl#6}[0]::caller[0]::DUMMY_LOCATION[0]
[Crux]     internal: static_initializer

[Crux-MIR] ---- FINAL RESULTS ----
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
[Crux] Overall status: Invalid.
```

Quite surprising, as I would have thought that https://github.com/GaloisInc/mir-json/pull/302/commits/fec37b1d2ae333dcbe66384fe6d6a99dd7ac220f would have been the _less_ disruptive change. I'll have to investigate to see what goes wrong here.

RyanGlScott:
Oh, this is because I made a very subtle typo in `try_render_ref_opty`. I changed `rty` to `op_ty.layout.ty`, but this wrong, as `op_ty` represents the reference type, not the type that is being pointed to. I should have used `d.layout.ty` instead.

I've fixed this in a force-push and verified that the `crux-test{,-comp}` test suites pass with this change.

RyanGlScott:
Superseded by #304.

PR #303, opened 2026-08-06
Implement Symbolic for Box, Option, Result, f16, f32, f64 and f128

### LINKED ISSUE #301: Implement Symbolic trait for Box and for Option
Both `Option` and `Box` are currently missing the respective `Symbolic` implementation. This should be fairly straightforward to implement, will help deriving `Symbolic` for some more complex data structures.

podhrmic:
Fixes #301

### DIFF
--- a/libs/Patches.md
+++ b/libs/Patches.md
@@ -341,6 +341,10 @@ into the main commit for that patch, and then the *Update* line can be removed.
   with the `unsupported` configuration, where all parking-related functions are
   treated as no-ops.
 
+* Implement `Symbolic` trait for `Box` (last applied: August 6, 2026)
+  Because the `alloc` crate depends on the `crucible` crate, we implement it
+  by patching the `alloc` crate.
+
 # Notes
 
 This section contains more detailed notes about why certain patches are written
--- a/libs/Patches.md
+++ b/libs/Patches.md
@@ -342,6 +342,7 @@ into the main commit for that patch, and then the *Update* line can be removed.
   treated as no-ops.
 
 * Implement `Symbolic` trait for `Box` (last applied: August 6, 2026)
+
   Because the `alloc` crate depends on the `crucible` crate, we implement it
   by patching the `alloc` crate.
 
--- a/libs/alloc/src/boxed.rs
+++ b/libs/alloc/src/boxed.rs
@@ -265,6 +265,13 @@ pub fn box_assume_init_into_vec_unsafe<T, const N: usize>(
     unsafe { (b.assume_init() as Box<[T]>).into_vec() }
 }
 
+#[stable(feature = "crucible", since = "1.92.0")]
+impl<T: crucible::Symbolic> crucible::Symbolic for Box<T> {
+    fn symbolic(desc: &str) -> Self {
+        Box::new(T::symbolic(desc))
+    }
+}
+
 impl<T> Box<T> {
     /// Allocates memory on the heap and then places `x` into it.
     ///
--- a/libs/alloc/src/boxed.rs
+++ b/libs/alloc/src/boxed.rs
@@ -265,13 +265,6 @@ pub fn box_assume_init_into_vec_unsafe<T, const N: usize>(
     unsafe { (b.assume_init() as Box<[T]>).into_vec() }
 }
 
-#[stable(feature = "crucible", since = "1.92.0")]
-impl<T: crucible::Symbolic> crucible::Symbolic for Box<T> {
-    fn symbolic(desc: &str) -> Self {
-        Box::new(T::symbolic(desc))
-    }
-}
-
 impl<T> Box<T> {
     /// Allocates memory on the heap and then places `x` into it.
     ///
--- a/libs/crucible/lib.rs
+++ b/libs/crucible/lib.rs
@@ -132,7 +132,7 @@ pub fn concretize<T>(x: T) -> T {
 }
 
 /// Install `g` as an override for `f`.
-pub fn override_<F, G>(f: F, g: G) {
+pub fn override_<F, G>(_f: F, _g: G) {
     unimplemented!("crucible::override_");
 }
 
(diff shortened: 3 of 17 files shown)


libs/crucible/symbolic.rs
```diff
@@ -39,6 +39,8 @@ core_impls! {
     u32, symbolic_u32;
     u64, symbolic_u64;
     u128, symbolic_u128;
+    f32, symbolic_f32;
+    f64, symbolic_f64;
```
RyanGlScott, line 44: Since you're in the neighborhood, could you also add instances for `f16` and `f128`? This would achieve parity with Kani and cover all of the primitive floating-point types that Rust offers.

libs/alloc/src/boxed.rs
```diff
@@ -265,6 +265,13 @@ pub fn box_assume_init_into_vec_unsafe<T, const N: usize>(
     unsafe { (b.assume_init() as Box<[T]>).into_vec() }
 }
 
+#[stable(feature = "crucible", since = "1.92.0")]
+impl<T: crucible::Symbolic> crucible::Symbolic for Box<T> {
+    fn symbolic(desc: &str) -> Self {
+        Box::new(T::symbolic(desc))
+    }
+}
```
RyanGlScott, line 273: Would you be willing to move this impl to `libs/crucible/symbolic.rs`? I like the idea of keeping all of the `Symbolic` impls in one file to make it easier to tell at a glance how many impls we define. (If there are technical reasons to keep the impl here, however, I'd be interested to hear them.)

libs/crucible/symbolic.rs
podhrmic: I am happy to! Turns out `f16` and `f128` are unstable, but:

```
#![feature(f16)]
#![feature(f128)]
```

seems to do the job.

libs/alloc/src/boxed.rs
podhrmic: No problem! The issue I am running into is how to import `Box` trait into `libs/crucible/symbolic.rs`:

```
 --> ./libs/crucible/symbolic.rs:3:19
  |
3 | use crate::alloc::boxed::Box;
  |                   ^^^^^ could not find `boxed` in `alloc`
```

I tried different variations of this, yet I am not quite sure how to reference the `alloc` crate from `crucible`

podhrmic:
Also, I will fix the library warnings, as they make it difficult to debug the output from translating libs (will be a separate commit)

libs/alloc/src/boxed.rs
RyanGlScott: Ah, my apologies, I overlooked the fact that there are technical reasons at play. Specifically, the `alloc` crate depends on the `crucible` crate, but I proposed code that would only work with the opposite dependency order. Reversing this order would be infeasible, so I retract my suggestion.

That being said, this does require that we patch the `core` crate to add code that wasn't there before, so we will need to remember to re-apply this patch on future Rust toolchain upgrades. Can you add an entry to [`libs/Patches.md`](https://github.com/GaloisInc/mir-json/blob/58261cbf47399acb31e8264df4a5cf0414cccf3e/libs/Patches.md) to record the existence of this patch?

libs/crucible/lib.rs
```diff
@@ -6,6 +6,8 @@
 #![feature(tuple_trait)]
 #![feature(type_alias_impl_trait)]
 #![feature(unboxed_closures)]
+#![feature(f16)]
+#![feature(f128)]
```
RyanGlScott: In keeping with the existing conventions in this file, can you put these `feature` lines in alphabetical order with the existing ones?

libs/alloc/src/boxed.rs
spernsteiner: >require that we patch the `core` crate

The `alloc` crate*

When doing this, please rebase such that the `alloc` changes and `Patches.md` changes are in the same commit, since this makes it easier to cherry-pick after future toolchain upgrades (you can `git blame lib/Patches.md`, find the commit that added the description of the patch you want, and cherry-pick it to get the patch itself).

podhrmic: Good point - done!

podhrmic:
In a similar vein - do you want me to add an entry into `Patches.md` about the warning fixes (unused variables etc) as that is something that may have to be redone in the future? (and rebase the commits accordingly)

libs/Patches.md
```diff
@@ -341,6 +341,10 @@ into the main commit for that patch, and then the *Update* line can be removed.
   with the `unsupported` configuration, where all parking-related functions are
   treated as no-ops.
 
+* Implement `Symbolic` trait for `Box` (last applied: August 6, 2026)
+  Because the `alloc` crate depends on the `crucible` crate, we implement it
```
RyanGlScott, line 346: For consistency with the rest of the file:


```suggestion
* Implement `Symbolic` trait for `Box` (last applied: August 6, 2026)

  Because the `alloc` crate depends on the `crucible` crate, we implement it
```

RyanGlScott (approved):

RyanGlScott:
> In a similar vein - do you want me to add an entry into `Patches.md` about the warning fixes (unused variables etc) as that is something that may have to be redone in the future?

`Patches.md` only concerns itself with code that was obtained upstream from Rust's standard libraries, and which we therefore must regenerate on each Rust toolchain upgrade. If I'm not mistaken, all of the warning fixes concern the `crucible` crate (which is all custom code) or the `byteorder`/`int512` crates (which are not part of Rust's standard libraries, but we include anyway for historical reasons). None of that rises to the level of needing to be mentioned in `Patches.md`, in my opinion.

podhrmic:
To clarify, some small patches were done in other libraries too (like `byteorder`) - but the changes were rather trivial, and in fact this is something a LLM can easily do ("fix unused variable warnings") should we want to redo this in the future

RyanGlScott:
`byteorder` is one of those crates that's not part of the Rust standard library, but is included here in `mir-json` for historical reasons. It's fine to leave out changes to the `byteorder` crate in `Patches.md`, as we don't need to regenerate it on every Rust toolchain upgrade.

PR #304, opened 2026-08-07
Evaluate static initializers at compile time

### LINKED ISSUE #177: Always emit rendered constants for static items

yav:
Fixes #177 by implementing the plan outlined in the discussion on this issue. 

Claude helped quite a bit with this PR, so it's possible it made some stuff up---I don't see anything that's wrong, so as far as I understand the implementation matches the suggestions from the issue, but we should be fairly careful with the review of this PR, as I am very new to working with the `rustc` API, so it is quite possible I am misunderstanding some of the details.

Also note that this is on top of `rgs/constant-eval-refactoroing`

### DIFF
--- a/doc/mir-json-schema.ts
+++ b/doc/mir-json-schema.ts
@@ -5,7 +5,7 @@
 
 /// The MIR JSON format
 type MIR = {
-  version: 12,
+  version: 13,
   fns: Fn[],
   adts: Adt[],
   statics: Static[],
@@ -192,9 +192,11 @@ type Layout = {
 // -----------------------------------------------------------------------------
 
 /// Globals
+///
+/// Both ordinary and thread-local statics are represented uniformly with
+/// `kind: "constant"`.
 type Static =
     { kind: "constant", name: DefId, ty: Ty, mutable: boolean, rendered: ConstVal }
-  | { kind: "body",     name: DefId, ty: Ty, mutable: boolean }
 
 
 // -----------------------------------------------------------------------------
--- a/doc/mir-json-schema.ts
+++ b/doc/mir-json-schema.ts
@@ -282,7 +282,7 @@ type ConstVal =
   | { kind: "strbody", len: number, elements: number[] }
   | { kind: "struct",  fields: ConstVal[] }
   | { kind: "enum", variant: number, fields: ConstVal[] }
-  | { kind: "union" }
+  | { kind: "union", variant: number, val: ConstVal }
   | { kind: "fndef", def_id: DefId }
   | { kind: "static_ref", def_id: DefId }
   | { kind: "zst" }
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -947,10 +947,6 @@ fn emit_static(ms: &mut MirState, out: &mut impl JsonOutput, def_id: DefId) -> i
     emit_fn(ms, out, &name, None, mir)?;
     emit_static_decl(ms, out, &name, mir.return_ty(), tcx.is_mutable_static(def_id))?;
 
-    for (idx, mir) in tcx.promoted_mir(def_id).iter_enumerated() {
-        emit_promoted(ms, out, &name, idx, mir)?;
-    }
-
     Ok(())
 }
 
@@ -1144,31 +1140,6 @@ fn emit_instance<'tcx>(
     let mir = tcx.arena.alloc(mir);
     emit_fn(ms, out, &name, Some(ty_inst), mir)?;
 
-    if let ty::InstanceKind::Item(def_id) = ty_inst.def {
-        for (idx, mir) in tcx.promoted_mir(def_id).iter_enumerated() {
-            let mir = ty_inst.instantiate_mir_and_normalize_erasing_regions(
-                tcx,
-                ty::TypingEnv::fully_monomorphized(),
-                ty::EarlyBinder::bind(mir.clone()),
-            );
-            let mir = tcx.arena.alloc(mir);
-            emit_promoted(ms, out, &name, idx, mir)?;
-        }
-    }
-
-    Ok(())
-}
-
-fn emit_promoted<'tcx>(
-    ms: &mut MirState<'_, 'tcx>,
-    out: &mut impl JsonOutput,
-    parent: &str,
-    idx: mir::Promoted,
-    mir: &'tcx Body<'tcx>,
-) -> io::Result<()> {
-    let name = format!("{}::{{{{promoted}}}}[{}]", parent, idx.as_usize());
-    emit_fn(ms, out, &name, None, mir)?;
-    emit_static_decl(ms, out, &name, mir.return_ty(), false)?;
     Ok(())
 }
 
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -876,7 +876,7 @@ fn emit_trait<'tcx>(
 }
 
 
-/// Emit all statics defined in the current crate.
+/// Emit all statics defined in the current crate. Extern statics are not exported.
 fn emit_statics(ms: &mut MirState, out: &mut impl JsonOutput) -> io::Result<()> {
     let parts = ms.tcx.collect_and_partition_mono_items(());
     for cgu in parts.codegen_units {
@@ -909,49 +909,34 @@ const X: &[u64] = &[42];
 static Y: &[&[u64]] = &[&X];
 
 When compiling this code, a nested static is created for the `&X` portion of
-`Y`'s body.
-
-Nested statics are treated very specially in the compiler, and they are not
-given types like other static items have. This doesn't cause issues for the
-flat "bytes and pointers" memory representation that rustc uses for the
-contents of allocations, but it is incompatible with the current crucible-mir
-memory model. The crucible-mir memory model requires a structured, typed
-representation of values, which we can extract from a flat allocation only if
-we know the correct type to extract into. Because nested statics don't
-explicitly store their types anywhere, our only hope of supporting them would
-be to stumble upon a reference to the same allocation in some other context
-where we do know the expected type.
-
-Thankfully, we can get away without needing to emit any entries for nested
-static items. This is because nested statics don't actually exist in the body
-of a static item itself. Rather, they are created as a side effect of
-evaluating the body down to a ConstAllocation, and the final ConstAllocation
-may reference those nested statics. (See
+`Y`'s body. See
 https://github.com/GaloisInc/mir-json/issues/129#issuecomment-3387602092 for a
-more thorough exploration of how nested static items arise during compilation.)
-mir-json, on the other hand, emits the MIR *body*, and to the best of our
-understanding, the MIR body exists prior to the creation of any nested statics.
-As a result, the emitted JSON for the MIR body won't reference nested statics
-at all.
+more thorough exploration of how nested static items arise during compilation.
 
-Implementation-wise, this means that we simply skip over any `MonoItem` that
-corresponds to a `Static` item with `nested: true`.
+Nested statics are treated specially in the compiler, and they are not given
+types like ordinary static items have. This is a problem for our renderer,
+which needs a structured, typed representation of each value. Instead of
+emitting nested statics as top-level entries in the `statics` table, we
+render them inline at the reference site: when `try_render_ref_opty` finds
+that a pointer points at a nested-static allocation, it evaluates that
+static's initializer and interns it under an anonymous `{{alloc}}` name,
+using the surrounding pointer's pointee type to drive the render.
+
+As a result, we skip any `MonoItem` here that corresponds to a `Static` with
+`nested: true` — those items are reached (and emitted) through their
+referring static rather than in their own right.
 */
 
+/// Emit a single static as a `kind: "constant"` entry in the `statics`
+/// table, evaluating and rendering its initializer via
+/// [`render_static_body`].
 fn emit_static(ms: &mut MirState, out: &mut impl JsonOutput, def_id: DefId) -> io::Result<()> {
     let tcx = ms.tcx;
     let name = def_id_str(tcx, def_id);
+    let ty = tcx.type_of(def_id).instantiate_identity();
+    let is_mut = tcx.is_mutable_static(def_id);
 
-    // let mir = tcx.optimized_mir(def_id);
-    let mir = tcx.mir_for_ctfe(def_id);
-    emit_fn(ms, out, &name, None, mir)?;
-    // Add a new static declaration to `out.statics`.
-    let j = json!({
-        "name": name,
-        "ty": mir.return_ty().to_json(ms),
-        "mutable": tcx.is_mutable_static(def_id),
-        "kind": "body",
-    });
+    let j = render_static_body(ms, def_id, ty, name, is_mut);
     out.emit(EntryKind::Static, j)?;
     emit_new_defs(ms, out)
 }
--- a/src/analyz/mod.rs
+++ b/src/analyz/mod.rs
@@ -945,23 +945,11 @@ fn emit_static(ms: &mut MirState, out: &mut impl JsonOutput, def_id: DefId) -> i
     // let mir = tcx.optimized_mir(def_id);
     let mir = tcx.mir_for_ctfe(def_id);
     emit_fn(ms, out, &name, None, mir)?;
-    emit_static_decl(ms, out, &name, mir.return_ty(), tcx.is_mutable_static(def_id))?;
-
-    Ok(())
-}
-
-/// Add a new static declaration to `out.statics`.
-fn emit_static_decl<'tcx>(
-    ms: &mut MirState<'_, 'tcx>,
-    out: &mut impl JsonOutput,
-    name: &str,
-    ty: ty::Ty<'tcx>,
-    mutable: bool,
-) -> io::Result<()> {
+    // Add a new static declaration to `out.statics`.
     let j = json!({
         "name": name,
-        "ty": ty.to_json(ms),
-        "mutable": mutable,
+        "ty": mir.return_ty().to_json(ms),
+        "mutable": tcx.is_mutable_static(def_id),
         "kind": "body",
     });
     out.emit(EntryKind::Static, j)?;
--- a/src/analyz/ty_json.rs
+++ b/src/analyz/ty_json.rs
@@ -2,6 +2,7 @@ use rustc_abi::FieldsShape;
 use rustc_data_structures::stable_hasher::{HashStable, StableHasher};
 use rustc_hashes::Hash64;
 use rustc_hir as hir;
+use rustc_hir::def::DefKind;
 use rustc_hir::def_id::DefId;
 use rustc_hir::lang_items::LangItem;
 use rustc_index::{IndexVec, Idx};
@@ -1487,6 +1488,10 @@ fn try_render_opty_upvars<'tcx>(
     Some(upvar_vals)
 }
 
+/// Produce a `"kind": "constant"` JSON entry with all fields except `name`
+/// filled in. Callers are responsible for setting `name`: `intern_alloc`
+/// lets `AllocIntern::insert` mint an anonymous `{{alloc}}` id, while
+/// `render_static_body` supplies the static's real name.
 fn make_allocation_body<'tcx>(
     mir: &mut MirState<'_, 'tcx>,
     icx: &mut interpret::InterpCx<'tcx, RenderConstMachine<'tcx>>,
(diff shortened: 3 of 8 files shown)


doc/mir-json-schema.ts
```diff
@@ -192,9 +192,10 @@ type Layout = {
 // -----------------------------------------------------------------------------
 
 /// Globals
+///
+/// `rendered` is `null` for extern statics, which have no initializer.
 type Static =
-    { kind: "constant", name: DefId, ty: Ty, mutable: boolean, rendered: ConstVal }
-  | { kind: "body",     name: DefId, ty: Ty, mutable: boolean }
+    { kind: "constant", name: DefId, ty: Ty, mutable: boolean, thread_local: boolean, rendered: ConstVal | null }
```
RyanGlScott: Just to check my understanding, is `ConstVal | null` equivalent to writing `ConstVal?`? (Everything else in this file currently uses the latter convention.)

src/analyz/ty_json.rs
```diff
@@ -1628,8 +1634,42 @@ fn try_render_ref_opty<'tcx>(
...
-        interpret::GlobalAlloc::Static(def_id) =>
-            def_id.to_json(mir),
+        interpret::GlobalAlloc::Static(def_id) => {
+            if matches!(tcx.def_kind(def_id), DefKind::Static { nested: true, .. }) {
+                // Nested statics have no name of their own that downstream can
+                // refer to, so we handle them the same way as `GlobalAlloc::Memory`:
+                // render the initializer here and intern it under an anonymous name,
+                // using the surrounding pointer's pointee type.
+
+                // Rustc does not emit threadlocal nested statics.
```
RyanGlScott:
```suggestion
                // Rustc does not emit thread-local nested statics.
```

src/analyz/ty_json.rs
```diff
@@ -1628,8 +1634,42 @@ fn try_render_ref_opty<'tcx>(
...
+                );
+                let ty = op_ty.layout.ty;
+                let def_id_str = match mir.allocs.get(ca, ty) {
+                    Some(alloc_id) => alloc_id.to_owned(),
+                    None => {
+                        let body = make_allocation_body(mir, icx, &d, is_mut);
+                        mir.allocs.insert(tcx, ca, ty, body)
+                    }
+                };
+                def_id_str.to_json(mir)
```
RyanGlScott: This code is copy-pasted from the code in the `interpret::GlobalAlloc::Memory` case. It would be nice to deduplicate the two pieces of code.

src/analyz/ty_json.rs
```diff
@@ -1717,6 +1757,54 @@ fn try_render_ref_opty<'tcx>(
...
+    // paths converge at the interned ConstAllocation, so downstream keying
+    // on `(ConstAllocation, Ty)` (as `AllocIntern` does) still works.
+    let alloc_id = tcx.reserve_and_set_memory_alloc(ca);
+    let ptr = Pointer::new(CtfeProvenance::from(alloc_id), Size::ZERO);
+    let mpty = icx.ptr_to_mplace(ptr.into(), layout);
+
+    let mut j = make_allocation_body(mir, &mut icx, &mpty, is_mut);
+    j["name"] = name.into();
+    j["thread_local"] = is_thread_local.into();
+    j
```
RyanGlScott, line 1800: It seems a bit delicate to incremenetally update JSON map entries like this. I would prefer passing `name` and `is_thread_local` as arguments to `make_allocation_body` such that we construct the JSON value all at once.

src/analyz/mod.rs
```diff
@@ -908,49 +918,50 @@ const X: &[u64] = &[42];
...
+/// Extern statics have no initializer available, so their `rendered`
+/// field is `null`; all other statics are evaluated and rendered via
+/// [`render_static_body`]. The `thread_local` flag is set uniformly
+/// either way.
 fn emit_static(ms: &mut MirState, out: &mut impl JsonOutput, def_id: DefId) -> io::Result<()> {
     let tcx = ms.tcx;
     let name = def_id_str(tcx, def_id);
+    let ty = tcx.type_of(def_id).instantiate_identity();
+    let is_mut = tcx.is_mutable_static(def_id);
+    let is_thread_local = tcx.is_thread_local_static(def_id);
```
RyanGlScott: Out of curiosity, does the presence or absence or `thread_local` affect how `crucible-mir`'s semantics should work? I ask since there are any number of attributes we could choose to record in the JSON, so I wonder why this one in particular is needed.

doc/mir-json-schema.ts
yav: No: `ConstVal | null` means that we either expect there to be a `ConstVal` (i.e., initial value) or `null` (i.e., we have an `extern` declaration which has no initializer).  I spent a bunch of time changing how to encode this and this way seemed pretty natural.  I am still unsure if we should have `thread_local` as a flag or another constructor, as imagine it is not very common, and like this we get a lot of `thread_local: false`...

RyanGlScott: I had assumed that `ConstVal?` was also equivalent to `ConstVal | null`. What does `ConstVal?` actually mean?

src/analyz/mod.rs
yav: I am not sure what we do in `crucible` but I think technically it matters because the thread local statics should have a different address, depending on the thread, unlike the global ones which have a single address.

src/analyz/mod.rs
```diff
@@ -894,6 +894,16 @@ fn emit_statics(ms: &mut MirState, out: &mut impl JsonOutput) -> io::Result<()>
             }
         }
     }
+
+    // Extern statics are not collected as mono items, so we
+    // enumerate them separately via the HIR.
```
spernsteiner: Why's it necessary to include these in the output?

src/analyz/mod.rs
spernsteiner: Why export this?  Nothing uses it, and we have no plans to use it either

src/analyz/ty_json.rs
```diff
@@ -1521,6 +1523,7 @@ fn make_allocation_body<'tcx>(
             icx: &mut interpret::InterpCx<'tcx, RenderConstMachine<'tcx>>,
             d: &MPlaceTy<'tcx>,
             len: u64,
+            is_mut: bool,
```
spernsteiner: Can this ever be `true` here, given `str` is unsized?

src/analyz/ty_json.rs
```diff
@@ -1628,8 +1631,42 @@ fn try_render_ref_opty<'tcx>(
...
+                );
+
+                let ca = tcx.eval_static_initializer(def_id).unwrap_or_else(|e|
+                    tcx.dcx().span_bug(
+                        tcx.def_span(def_id),
+                        format!(
+                            "eval_static_initializer failed for nested static {:?}: {:?}",
+                            def_id, e,
+                        ),
+                    )
```
spernsteiner: Way overkill—we normally just panic instead of emitting proper rust diagnostics.  We assume the code already compiles, so operations like this should be guaranteed to succeed.

src/analyz/ty_json.rs
```diff
@@ -1717,6 +1754,54 @@ fn try_render_ref_opty<'tcx>(
...
+    );
+
+    let layout = tcx.layout_of(
+        ty::TypingEnv::fully_monomorphized().as_query_input(ty),
+    ).unwrap();
+
+    // Mint a fresh AllocId for this ConstAllocation. This is distinct from
+    // whatever AllocId rustc uses for this static internally, but the two
+    // paths converge at the interned ConstAllocation, so downstream keying
+    // on `(ConstAllocation, Ty)` (as `AllocIntern` does) still works.
```
spernsteiner: >the two paths converge at the interned ConstAllocation, so downstream keying on `(ConstAllocation, Ty)` still works.

AI slop.  What "two paths"?  As noted in this very same sentence, we don't use the `AllocId`s for anything, only the `ConstAllocation`, which is interned.

doc/mir-json-schema.ts
yav: Oh, sorry I didn't see that there were two question marks.  Yes, `ConstVal?` should be syntactic sugar for `ConstVal | null`.  I can change it to use the question mark

doc/mir-json-schema.ts
```diff
@@ -5,7 +5,7 @@
 
 /// The MIR JSON format
 type MIR = {
-  version: 12,
+  version: 13,
```
spernsteiner, line 8: Looks like this got added to the wrong commit—the schema changes and changelog update were in the previous one.

src/analyz/ty_json.rs
yav: Good point: I changed it to pass the `thread_local` flag.  The `name`, however, is not know for the anonymous case and is generated when we insert it in the table:

https://github.com/GaloisInc/mir-json/blob/1b2566b265fd22dd6788d990886b3e7be8971557/src/analyz/to_json.rs#L417

So I think it's probably fine to set it manually here.

src/analyz/ty_json.rs
```diff
@@ -1510,6 +1510,7 @@ fn make_allocation_body<'tcx>(
         json!({
             "kind": "constant",
             "mutable": is_mut,
+            "thread_local": false,
```
spernsteiner: Actually this whole commit should probably be folded into the previous one, since `thread_local` fields were already added there in some places.

src/analyz/ty_json.rs
```diff
@@ -1525,6 +1531,7 @@ fn make_allocation_body<'tcx>(
             d: &MPlaceTy<'tcx>,
             len: u64,
             is_mut: bool,
+            is_thread_local: bool,
```
spernsteiner: Again, I don't think this can ever be true

src/analyz/ty_json.rs
```diff
@@ -1802,9 +1810,8 @@ pub fn render_static_body<'tcx>(
     let ptr = Pointer::new(CtfeProvenance::from(alloc_id), Size::ZERO);
     let mpty = icx.ptr_to_mplace(ptr.into(), layout);
 
-    let mut j = make_allocation_body(mir, &mut icx, &mpty, is_mut);
+    let mut j = make_allocation_body(mir, &mut icx, &mpty, is_mut, is_thread_local);
     j["name"] = name.into();
-    j["thread_local"] = is_thread_local.into();
     j
```
spernsteiner: Old version seems better, since we're already doing this sort of post-call update for the `name` field, and `thread_local: true` can appear only in very limited cases.

src/analyz/mod.rs
yav: Well, it is one of the things the Rust compiler can produce, and `mir-json` is supposed to export MIT to JSON.  I imagine it's not super common, but it did seem like we could actually support this one day (e.g., as a symbolic value).  I suppose to be useful we'd need some sort of way to declare an invariant on it, but is there a reason to *not* export them?

src/analyz/mod.rs
yav: Do you mean the `thread_local` flag?  I thought it'd be nice to know which ones are really unique and which ones are thread_local as the thread local ones would have a different address, which depends on the thread.  I imagine we don't support this yet, but I thought since I am looking into it now, I might as well do it. 

I did go back and forth on the representation though, and I am not sure fi we should have it as a field or anotehr constructor, as it fairly uncommon I imagine.

src/analyz/ty_json.rs
yav: Since I don't fully understand what the invariants on things are, I figured it's better to be a bit more verbose.   I can change it, what's a better way to do it?

src/analyz/ty_json.rs
yav: heh, actually that was my "correction" of Claude's original more verbose comment.  What's a better way to say this?  I wanted a brief way to say the fact that we make a bogus allocation id doesn't matter because we are just interested in the content, and the keys would be correct anyway.  At least, that's my understanding of what's going on.

doc/mir-json-schema.ts
yav: Yes, I forgot to update the schema in the spec.   It might be better to look at the final thing rather than the individual commits, especially since I've been making lot's of little tweaks based on the comments.

src/analyz/ty_json.rs
yav: Yes, the commits are wonky, I was fixing thing, sorry about that.  If it's an issue I can squash them in a single commit once we are happy with what the code looks like.

src/analyz/mod.rs
spernsteiner: Just adds complexity for something that AFAIK we'll likely never have a use for.  Any code using extern statics is likely also using extern fns that we can't support.

src/analyz/mod.rs
spernsteiner: The crux-mir multithreading support is pretty bitrotted, so we can't actually make use of this information right now, and it seems like it's adding a good bit of noise to the diff (extra args in several places).  IMO it would be better to just leave it out.  This is similar to how we don't bother exporting any fields for types or statement/terminator kinds that crux-mir can't handle.

src/analyz/ty_json.rs
spernsteiner: I mean the fancy `dcx().span_bug(...)` error reporting is overkill.  This should `panic!()`, or maybe even just `unwrap()` with no special message.  (And similarly for the other use of `dcx` below.)

src/analyz/ty_json.rs
spernsteiner:
```suggestion
    // Mint a fresh AllocId for this ConstAllocation. This is distinct from
    // whatever AllocId rustc uses for this static internally, but this is okay
    // because we only use the interned ConstAllocation and not the AllocId
    // to identify the allocation (such as in AllocIntern keys).
```

src/analyz/mod.rs
yav: I ended up changing the schema to use kind `thread-local` instead, so most of the generated `mir` files should end up looking like before (except always using `rendered` for the initializer).   I also removed the `thread-local` parameters from the functions as with this change we just need to do a single tweak to the constructor at the one place where we emit the local, which does not seem that invasive.

I do think it's worth noting the difference to avoid surprises if we ever resurrect multi threading support, but if we decide to remove it it should be just a few line change (+ update the schema)

RyanGlScott: I'm inclined to agree with @spernsteiner that we should just leave this out unless we have a specific need for it. My suspicion is that we'd need more information than just `thread_local` to do multi-threading correctly, and without a way to test this on the `crucible-mir` side, it's hard to validate that what we have in this PR is enough.

doc/mir-json-schema.ts
```diff
@@ -192,9 +192,11 @@ type Layout = {
 // -----------------------------------------------------------------------------
 
 /// Globals
+///
+/// Both ordinary and thread-local statics are represented uniformly with
+/// `kind: "constant"`.
```
RyanGlScott, line 197: The comments about thread-local statics are now out of date.

RyanGlScott (commented):
LGTM, aside from one minor thing.

doc/mir-json-schema.ts
yav: This comment seems accurate, or am I misunderstanding?

yav: I did notice late on Friday that the schema is inaccurate, in that the initializer can be null if we fail to export the constant. I am not sure if it's better to update the schema, or make mir-json panic in that case.

RyanGlScott: Apologies, I misunderstood what exactly changed. I was (mistakenly) under the impression that `mir-json` now ignored thread-local statics, but now that I've re-read this, that's not true---it exports thread-local statics as `"constant"` as well.

> the initializer can be null if we fail to export the constant.

Can you give an example of where this happens?

yav: It's because I used `try_render_opty`.  I just saw that we have a non try version, which handles the `none` case and emits `unsupported`, which is probably what I should have used

RyanGlScott (approved):
It would be worth testing this in `crux-mir`'s test suite first, but otherwise this LGTM.

src/analyz/ty_json.rs
```diff
@@ -1236,29 +1236,48 @@ pub fn render_opty<'tcx>(
...
-        json!({
-            "kind": "unsupported_const",
-            "debug_val": format!("{:?}", op_ty),
-        })
+    try_render_opty(mir, icx, op_ty).unwrap_or_else(|err| {
+        match err {
+            RenderErr::Mismatch => panic!("Unexpexpected type mismatch"),
+            RenderErr::Unsupported =>
+                json!({
+                    "kind": "unsupported_const",
```
RyanGlScott: This is somewhat out of sync with the JSON schema, which describes unsupported `ConstVal`s as having kind `unsupported`:

https://github.com/GaloisInc/mir-json/blob/2a3542ca08abb8ad6f35a1ae29b534cfaf112cb6/doc/mir-json-schema.ts#L295

We should either change the code here to be in sync with the schema, or we should update the schema to use `unsupported_const`.

src/analyz/ty_json.rs
```diff
@@ -1464,10 +1507,10 @@ pub fn try_render_opty<'tcx>(
         ty::TyKind::Pat(ty, _) => {
             let inner_ty_and_layout = mir.tcx.layout_of(
                 ty::TypingEnv::fully_monomorphized().as_query_input(ty)).unwrap();
-            let new_op_ty = op_ty.transmute(inner_ty_and_layout, icx).unwrap();
+            let new_op_ty = check_mismatch(op_ty.transmute(inner_ty_and_layout, icx))?;
             return try_render_opty(mir, icx, &new_op_ty);
         },
-        ty::TyKind::UnsafeBinder(_unsafe_binder_inner) => {
+        ty::TyKind::UnsafeBinder(_unsafe_binder_inner) => {  // XXX: is this differnt from Err::Unsupported?
```
RyanGlScott: I think your comment is spot-on: we ought to be using `Err(RenderErr::Unsupported)` here for consistency.

src/analyz/ty_json.rs
```diff
@@ -1236,29 +1236,48 @@ pub fn render_opty<'tcx>(
...
+                json!({
+                    "kind": "unsupported_const",
+                    "debug_val": format!("{:?}", op_ty),
+                })
+        }
     })
 }
 
+pub enum RenderErr { Mismatch, Unsupported }
+pub type Partial<T> = Result<T, RenderErr>;
```
RyanGlScott: Can you add some documentation to `RenderErr` and `Partial`? In particular, it would be worth noting that `Mismatch` indicates a _type_ mismatch.

src/analyz/ty_json.rs
```diff
@@ -1354,7 +1374,31 @@ pub fn try_render_opty<'tcx>(
...
+                        Err(RenderErr::Unsupported) => return Err(RenderErr::Unsupported)
+                    }
+                }
+                match field_val {
+                  None => return Err(RenderErr::Mismatch),
+                  Some((idx,_,val)) => {
+                    let js: serde_json::Value = val.into();
+                    json!({"kind": "union", "variant": idx.as_u32(), "val": js })
+                  }
+                }
```
RyanGlScott, line 1411: The indentation here is inconsistent. I'd prefer uniformly using four spaces to indent, as is the convention in most Rust codebases:


```suggestion
                for field_idx in variant.fields.indices() {
                    let field = check_mismatch(icx.project_field(op_ty, field_idx))?;
                    let sz    = field.layout.size;

                    // We are looking for the largest interpretation that works
                    // If there are multiple with the same size we pick the
                    // first one.
                    if let Some((_,cur_sz,_)) = field_val {
                        if cur_sz >= sz { continue }
                    }
                    match try_render_opty(mir, icx, &field) {
                        Ok(v) => field_val = Some((field_idx,sz,v)),
                        Err(RenderErr::Mismatch) => (),
                        Err(RenderErr::Unsupported) => return Err(RenderErr::Unsupported)
                    }
                }
                match field_val {
                    None => return Err(RenderErr::Mismatch),
                    Some((idx,_,val)) => {
                        let js: serde_json::Value = val.into();
                        json!({"kind": "union", "variant": idx.as_u32(), "val": js })
                    }
                }
```

And similarly elsewhere.

RyanGlScott (commented):
I think there was a botched merge at some point, since this PR's diff is showing changes from https://github.com/GaloisInc/mir-json/pull/312.

RyanGlScott (approved):

PR #309, opened 2026-08-13
`libs`: Remove patches made obsolete by `MirAggregate` flattening

### LINKED ISSUE #293: libs: remove `core::slice::crucible_cast_hook`
Once https://github.com/GaloisInc/crucible/pull/1842 is merged, this hook will no longer need to be defined, as `crucible-mir` will be able to simulate the actual computation it shadows, and so will no longer target this hook.

RyanGlScott:
Now that `MirAggregate` flattening has landed, many of `mir-jsons`'s standard library patches are no longer necessary, as `crucible-mir` can now simulate many more pointer casts than it could before. This PR removes as many as possible.

Two patches which I considered removing (but ultimately abstained) are:

* https://github.com/GaloisInc/mir-json/pull/277/changes/9a13561525ec45637c0fa3b6a7a2c9cffaba7b30 (`libs: use hook in core::array::from_ref`)
* https://github.com/GaloisInc/mir-json/pull/277/changes/fee07c8755b28bfbe7152384ea076d7559d90320 (`libs: use hooks in core::slice::from_ref/mut`)

These functions use pointer casts from `*const T` to `*const [T; 1]`, and some testing suggests that `crucible-mir` doesn't support these types of casts just yet. (Should it?) 

Fixes #293.

### DIFF
--- a/libs/Patches.md
+++ b/libs/Patches.md
@@ -258,16 +258,6 @@ into the main commit for that patch, and then the *Update* line can be removed.
   back to using the global allocator instead.  We already don't support custom
   global allocators, so this shouldn't cause any problems.
 
-* Add `{Arc,Rc}::{from,into}_inner_raw` API functions (last applied: June 12, 2026)
-
-  `crucible-mir` is not currently able to simulate the
-  `{Arc,Rc}::{from,into}_raw` functions, as they rely on non-trivial pointer
-  offsets plus read from type-casted pointers.
-  `{Arc,Rc}::{from,into}_inner_raw` offer alternatives with very similar types,
-  but which are implemented in a Crucible-friendly way.
-
-  See also the "`{Arc,Rc}::{from,into}_inner_raw`" below.
-
 * Don't use a `union` in `LazyLock`'s internals (last applied June 12, 2026)
 
   The internals of `std::sync::LazyLock` use a `union` value to distinguish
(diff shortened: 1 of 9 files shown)


spernsteiner:
These commits all say that they partially revert various library commits, but they only remove things from `Patches.md` without adding or updating anything.  My previous workflow for toolchain upgrades was to `git blame libs/Patches.md` and then cherry-pick the commit that added the patch description there, since that commit also contains the relevant code changes.  If this branch is making code changes without adding anything to `Patches.md`, then those changes would be lost if the next upgrade uses the same workflow.

1. If any of these are fully reverting a commit (instead of reverting only parts of it), please update the messages to make that clear.
2. For partial reverts, it might be better to do a full revert and then apply a new, revised patch as a separate commit.  That way `git blame libs/Patches.md` will point only to the latest version of the patch.  Updating the "last applied" line would help but wouldn't be as convenient, since during the toolchain upgrade we'd need to cherry-pick two different commits, and if we ever made multiple edits to the same patch, only the last one would be visible in `git blame`.

RyanGlScott:
The reason I used the phrasing "partially reverts" in the commit message for https://github.com/GaloisInc/mir-json/pull/309/changes/d7cd9cfd5b1d61232659e0c956b467cafa1d30d3 is because there is exactly one change from https://github.com/GaloisInc/mir-json/commit/9dbf1ee3b99af9712627db40f559da70a9077962 that it does not revert. Namely https://github.com/GaloisInc/mir-json/commit/9dbf1ee3b99af9712627db40f559da70a9077962 adds the `` Mark hook functions as `#[inline(never)]` `` section to `Patches.md`. This section is still relevant even this PR, as there are other hook functions besides the ones removed in https://github.com/GaloisInc/mir-json/commit/9dbf1ee3b99af9712627db40f559da70a9077962. I'll update the PR to do a full revert of commit https://github.com/GaloisInc/mir-json/commit/9dbf1ee3b99af9712627db40f559da70a9077962 and then add the `` Mark hook functions as `#[inline(never)]` `` section back in a separate commit.

Upon closer inspection, I shouldn't have used the phrasing "partially reverts" in the other commits in this PR, as they are actually full reversions. I'll update the commit messages for those commits as well.

spernsteiner (approved):
Looks reasonable to me.  I believe we now have at least one crux-mir test for every patch, so Crucible CI should catch if any of the removed patches were actually needed.

PR #315, opened 2026-08-25
Do more validation on values (e.g., valid bool, valid char)

### LINKED ISSUE #314: More validation when rendering static constants
Upon further review of the changes from #177, @spernsteiner and I decided that it'd be good to do the following updates:

  * Don't panic in`render_opty`, instead handle like unsupported.  The reason for this is that it would appear that by using unions in Rust one can create all kinds of odd examples that would trigger a panic (mostly because there seem to be few guarantees about the possible values inhabiting a union type), and by using `unsupported` we can at least export the code, which could be verified, as long as the offending constant is not used
  * Add more validation when creating `bool` and `char` values. Currently we accept any value of the right size, but those have specific values that are accepted (e.g., 0 and 1 for `bool`).  Invalid values should be a `mismatch`
  *  Along the same lines as the previous bullets, we could do a bit more validation on ref types (e.g., that the number is suitably aligned for the given type, and there's enough space after the number to accommodate the type)
  * A few cases where we use `todo!`, should be `unsupported`
  * We should treat the distinction between function and data pointers as a `mismatch` rather then `unsupported`.

yav:
Also, we update some more of the error behaviors---turn things into explicit "unsupported", and don't panic at the top level if we find a mismatch, but just report unsupported.

Fixes #314

### DIFF
--- a/src/analyz/ty_json.rs
+++ b/src/analyz/ty_json.rs
@@ -1499,8 +1499,8 @@ pub fn try_render_opty<'tcx>(
             })
         }
 
-        ty::TyKind::Coroutine(_, _) => return Err(RenderErr::Unsupported), // not supported in haskell
-        ty::TyKind::CoroutineWitness(_, _) => return Err(RenderErr::Unsupported), // not supported in haskell
+        ty::TyKind::Coroutine(_, _) => return Err(RenderErr::Unsupported),
+        ty::TyKind::CoroutineWitness(_, _) => return Err(RenderErr::Unsupported),
 
         ty::TyKind::Tuple(elts) => {
             let mut vals = Vec::with_capacity(elts.len());
--- a/src/analyz/ty_json.rs
+++ b/src/analyz/ty_json.rs
@@ -1236,25 +1236,28 @@ pub fn render_opty<'tcx>(
     icx: &mut interpret::InterpCx<'tcx, RenderConstMachine<'tcx>>,
     op_ty: &interpret::OpTy<'tcx>,
 ) -> serde_json::Value {
-    try_render_opty(mir, icx, op_ty).unwrap_or_else(|err| {
-        match err {
-            RenderErr::Mismatch => panic!("Unexpexpected type mismatch"),
-            RenderErr::Unsupported =>
-                json!({
-                    "kind": "unsupported",
-                    "debug_val": format!("{:?}", op_ty),
-                })
-        }
+    try_render_opty(mir, icx, op_ty).unwrap_or_else(|_err| {
+        // We report "unsupported" even when we encounter "mismatch".  The reason
+        // is that the constraints on what can be stored in unions in Rust
+        // is pretty lax at the moment, but we actually look for well-formed
+        // values in the union alternatives.  So if we encounter one of the
+        // unusual union values (e.g., half one side of the union, half the
+        // other, which apparently is fine in stable Rust at the moment),
+        // we may get a mismatch.
+        json!({
+            "kind": "unsupported",
+            "debug_val": format!("{:?}", op_ty),
+        })
     })
 }
 
 
 /// Errors that may occur while interpreting bytes as a value of a specific type
 pub enum RenderErr {
-    /// Indicates that the bytes do not match the given type
+    /// Indicates that the bytes do not match the given type.
     Mismatch,
 
-    /// Indicates that we encountered a type that we don't know how to process
+    /// Indicates that we encountered a type that we don't know how to process.
     Unsupported
 }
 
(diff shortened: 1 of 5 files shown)


src/analyz/ty_json.rs
```diff
@@ -1236,25 +1236,28 @@ pub fn render_opty<'tcx>(
...
-                })
-        }
+    try_render_opty(mir, icx, op_ty).unwrap_or_else(|_err| {
+        // We report "unsupported" even when we encounter "mismatch".  The reason
+        // is that the constraints on what can be stored in unions in Rust
+        // is pretty lax at the moment, but we actually look for well-formed
+        // values in the union alternatives.  So if we encounter one of the
+        // unusual union values (e.g., half one side of the union, half the
+        // other, which apparently is fine in stable Rust at the moment),
+        // we may get a mismatch.
```
spernsteiner, line 1246: Should mention `Mix`/`Z0` so the reader can look at a concrete example of this "half-and-half" state.

src/analyz/ty_json.rs
```diff
@@ -1283,27 +1294,31 @@ pub fn try_render_opty<'tcx>(
...
 
     Ok(match *ty.kind() {
-        ty::TyKind::Bool |
-        ty::TyKind::Char |
-        ty::TyKind::Uint(_) =>
+        ty::TyKind::Bool => {
+            let s = check_mismatch(icx.read_immediate(op_ty))?.to_scalar();
+            let size = layout.size();
+            let bits = s.to_bits(size).unwrap();
+            if !(bits == 0 || bits == 1) { return Err(RenderErr::Mismatch) }
```
spernsteiner: `Scalar` actually has `to_char()` and `to_bool()` methods that fail (in `InterpResult`) if the value is no good.

src/analyz/ty_json.rs
```diff
@@ -1412,8 +1415,10 @@ pub fn try_render_opty<'tcx>(
             },
         },
 
-        ty::TyKind::Foreign(_) => todo!("foreign is unimplemented"), // can't do this
+        ty::TyKind::Foreign(_) => return Err(RenderErr::Unsupported), // can't do this
```
spernsteiner:
```suggestion
        // `Foreign` probably can't occur here, since it's unsized
        ty::TyKind::Foreign(_) => return Err(RenderErr::Unsupported),
```
Just clarifying the comment slightly

src/analyz/ty_json.rs
```diff
@@ -1739,8 +1747,9 @@ fn try_render_ref_opty<'tcx>(
             return raw_ptr(offset);
         }
         _ =>
-            // Give up
-            return Err(RenderErr::Unsupported)
+            // Give up.  We return a Mismatch, because we exepected a data
+            // pointer but we got a function or vtable one.
+            return Err(RenderErr::Mismatch)
```
spernsteiner:
```suggestion
            // Give up.  We return a Mismatch, because we expected a data
            // pointer but we got a function or vtable one.
            return Err(RenderErr::Mismatch)
```

tests/regression/test_statics/test.rs
```diff
@@ -43,8 +43,50 @@ pub static mut MUT_ARR: [u32; 3] = [1, 2, 3];
...
 
+pub union Mix {
+  f1: (bool, u8),
+  f2: (u8, bool),
+}
+
+// Z0 has a value which is half from one of the union and half from the other
+pub static Z0: Mix = {
+    let mut u = Mix { f1: (true, 99) };
+    u.f2.0 = 99; // overwrite through *the other* union, but only half the value
```
spernsteiner:
```suggestion
    u.f2.0 = 99; // overwrite through *the other* union variant, but only half the value
```

tests/regression/test_statics/test.rs
```diff
@@ -43,8 +43,50 @@ pub static mut MUT_ARR: [u32; 3] = [1, 2, 3];
...
+   let f1 = f as fn(i32) -> i32;
+   T(unsafe { core::mem::transmute(f1) })
+};
+
+
+pub union U { cast: fn(u32) -> u32 }
+
+// We can use a union to circumvent value expectations:
+// here's how we can make a function pointer, which has the value 8.
+// Without the union `rust` reports "undefined behavior".
```
spernsteiner:
```suggestion
// Without the union `rustc` reports "undefined behavior".
```

tests/regression/test_statics/test.rs
```diff
@@ -43,8 +43,50 @@ pub static mut MUT_ARR: [u32; 3] = [1, 2, 3];
...
+// Some function, so we can initialize function pointers.
+fn f(arg: i32) -> i32 { arg }
+
+// `Z1` points to `f`
+pub static Z1: fn(i32) -> i32 = f;
+
+pub struct T(*const());
+unsafe impl Sync for T {}
+
+// `Z2` is a const*() pointer that points to a function
```
spernsteiner:
```suggestion
// `Z2` is a *const() pointer that points to a function
```

spernsteiner (approved):

src/analyz/ty_json.rs
```diff
@@ -1489,8 +1498,9 @@ pub fn try_render_opty<'tcx>(
                 "upvars": upvar_vals,
             })
         }
-        ty::TyKind::Coroutine(_, _) => todo!("coroutine not supported yet"), // not supported in haskell
-        ty::TyKind::CoroutineWitness(_, _) => todo!("coroutinewitness not supported yet"), // not supported in haskell
+
+        ty::TyKind::Coroutine(_, _) => return Err(RenderErr::Unsupported), // not supported in haskell
+        ty::TyKind::CoroutineWitness(_, _) => return Err(RenderErr::Unsupported), // not supported in haskell
```
RyanGlScott: Not your fault, but I find the "not supported in haskell" comments here strange. Perhaps this should say "not supported in crucible-mir" instead?

RyanGlScott (approved):

Issue #298, opened 2026-07-30
`libs`: Introduce `BoundedSymbolic` trait


RyanGlScott:
The `Symbolic` trait is useful for quickly creating symbolic values of many different types, but it has its limitations. One particularly notable limitation is that it is unclear how to define `Symbolic` impls for non-fixed-sized types such as `Vec`. While it is clear how one would create symbolic `Vec` elements, it is much less clear what size the overall `Vec` should be. Zero elements? One elements? More? Restricting a `Vec` to a certain size means that you won't explore code paths where `Vec`s have larger sizes, and having the `Vec` be a symbolic size is also likely not what you want, given that Crucible's performance will tank if you try to simulate a symbolically-sized `Vec`.

On the other hand, it is annoying when you can't derive `Symbolic` impls just because one of the fields happens to be a non-fixed-sized type like `Vec`. As a practical consideration, it would be nice to be able to quickly derive _some_ sort of a way to create symbolic values of these types, even if we have to limit the size of the non-fixed-types to some upper bound. Kani does this by way of the derivable [`BoundedArbitrary`](https://github.com/model-checking/kani/blob/3acb374144e3837138db3320e8ee275d9693a085/library/kani_core/src/bounded_arbitrary.rs#L12-L14) trait, which is defined like so:

```rs
pub trait BoundedArbitrary {
    fn bounded_any<const N: usize>() -> Self;
}
```

The type of `bounded_any` is much like the `Arbitrary` trait's [`any`](https://github.com/model-checking/kani/blob/3acb374144e3837138db3320e8ee275d9693a085/library/kani_core/src/arbitrary.rs#L26) method, except that `bounded_arbitrary` takes a constant upper bound `N`. This allows you to define impls for `Vec` [like so](https://github.com/model-checking/kani/blob/3acb374144e3837138db3320e8ee275d9693a085/library/kani/src/bounded_arbitrary.rs#L27-L35):

```rs
impl<T: Arbitrary> BoundedArbitrary for Vec<T> {
    fn bounded_any<const N: usize>() -> Self {
        let real_length = kani::any_where(|&size| size <= N);
        let array: [T; N] = kani::any();
        let mut vec = Vec::from(array);
        vec.truncate(real_length);
        vec
    }
}
```

[...]

PR #318, opened 2026-09-01
Add BoundedSymbolic trait

podhrmic:
This change adds support for bounded symbolic representations of `String` and `Vec` types.


libs/crucible/symbolic.rs
```diff
@@ -171,3 +171,13 @@ impl<T: Symbolic, E: Symbolic> Symbolic for Result<T, E> {
...
+
+pub trait BoundedSymbolic: Sized {
+    fn bounded_symbolic<const N: usize>(desc: &str) -> Self;
+
+    fn bounded_symbolic_where<const N: usize, F: FnOnce(&Self) -> bool>(desc: &str, f: F) -> Self {
+        let x = Self::bounded_symbolic::<N>(desc);
+        super::crucible_assume!(f(&x));
+        x
+    }
+}
```
RyanGlScott, line 187: We should include some docstrings here. (See the docstrings for `Symbolic` for inspiration.) In particular, we should document what `N` means.

libs/crucible/lib.rs
```diff
@@ -29,6 +29,7 @@ pub use self::alloc::TypedAllocator;
 
 // Re-export the `Symbolic` trait, which is used to create symbolic values.
 pub use self::symbolic::Symbolic;
+pub use self::symbolic::BoundedSymbolic;
```
RyanGlScott: The comment here needs to be updated:

```suggestion
// Re-export the `Symbolic` and `BoundedSymbolic` traits, which are used to create symbolic values.
pub use self::symbolic::Symbolic;
pub use self::symbolic::BoundedSymbolic;
```

It might also be worth combining these into a single `pub use self::symbolic::{Symbolic, BoundedSymbolic};` statement, depending on your tastes.

libs/alloc/src/vec/mod.rs
```diff
@@ -444,6 +444,18 @@ pub struct Vec<T, #[unstable(feature = "allocator_api", issue = "32838")] A: All
...
+impl<T: crucible::Symbolic> crucible::BoundedSymbolic for Vec<T> {
+    /// Create a new symbolic `Vec<T>` with possible lengths bounded by the specified constant.
+    fn bounded_symbolic<const N: usize>(desc: &str) -> Self {
+        let array = <[T; N] as crucible::Symbolic>::symbolic(desc);
+        let mut vec = Vec::from(array);
+        let n = <usize as crucible::Symbolic>::symbolic_where("n", |n| n <= &N);
+        vec.truncate(n);
+        vec
+    }
+}
```
RyanGlScott, line 457: We are updating the standard libraries as part of this changeset, so we will need to make a note of this in `libs/Patches.md` as a reminder to reapply this patch going forward. My recommendation would be to add something like the following underneath [this part of `Patches.md`](https://github.com/GaloisInc/mir-json/blob/8cbf9af1821075ef47cf784fef78d2650827730d/libs/Patches.md#L298-L301):

> *Update* (September 1, 2026): Also implement `BoundedSymbolic` for `Vec` and `String`.

libs/alloc/src/string.rs
```diff
@@ -412,6 +412,19 @@ pub struct FromUtf8Error {
 #[derive(Debug)]
 pub struct FromUtf16Error(());
 
+#[stable(feature = "crucible", since = "1.92.0")]
+impl crucible::BoundedSymbolic for String {
+    /// Create a new symbolic `String` of valid utf8 code points.
+    /// The possible lengths of the symbolic string are bounded by the specified constant.
```
RyanGlScott: It would be worth noting that in this impl, `N` denotes the number of bytes in the `String`, not the number of characters.

libs/alloc/src/string.rs
```diff
@@ -412,6 +412,19 @@ pub struct FromUtf8Error {
...
+    fn bounded_symbolic<const N: usize>(desc: &str) -> Self {
+        let vec = <Vec<u8> as crucible::BoundedSymbolic>::bounded_symbolic::<N>(desc);
+        // Uses optimized str::from_utf8
```
RyanGlScott: I'm not sure I understand what information this comment is trying to convey that isn't readily apparent from the fact that we're invoking the `str::from_utf8` function below. Can we just remove this comment?

RyanGlScott (changes requested):
Please mention that this fixes https://github.com/GaloisInc/mir-json/issues/298 in the commit message. (You may also want to add yourself as a co-author in the commit message.)

libs/alloc/src/vec/mod.rs
spernsteiner: When you do this, please rebase/amend so that the `Patches.md` change is part of the same commit as the library change.  This makes it easier to port the library changes to newer rust toolchains in the future.

libs/alloc/src/string.rs
```diff
@@ -412,6 +412,19 @@ pub struct FromUtf8Error {
...
+        let result = str::from_utf8(vec.as_slice());
+        crucible::crucible_assume!(result.is_ok());
+        result.unwrap().to_string()
```
spernsteiner: You can go directly from `Vec` to `String` using `String::from_utf8(vec)`

libs/alloc/src/string.rs
```diff
@@ -412,6 +412,19 @@ pub struct FromUtf8Error {
...
 pub struct FromUtf16Error(());
 
+#[stable(feature = "crucible", since = "1.92.0")]
+impl crucible::BoundedSymbolic for String {
+    /// Create a new symbolic `String` of valid utf8 code points.
+    /// The possible lengths of the symbolic string are bounded by the specified constant.
+    fn bounded_symbolic<const N: usize>(desc: &str) -> Self {
+        let vec = <Vec<u8> as crucible::BoundedSymbolic>::bounded_symbolic::<N>(desc);
+        // Uses optimized str::from_utf8
+        let result = str::from_utf8(vec.as_slice());
```
podhrmic: On second though, I think I would like to `crucible_assume!` that the array contains only ASCII characters. Unless of course I can constraint the resulting string to contain only ASCII, but I am not sure how to best do that.

RyanGlScott: Why ASCII only?

libs/crucible/symbolic.rs
AlexMahrer: Done.

libs/crucible/lib.rs
AlexMahrer: Done.

libs/alloc/src/vec/mod.rs
AlexMahrer: Done.

libs/alloc/src/string.rs
AlexMahrer: Done.

libs/alloc/src/string.rs
AlexMahrer: Removed.

libs/alloc/src/string.rs
AlexMahrer: Good point! This change has been made.

libs/alloc/src/string.rs
podhrmic: I am a bit worried about the performance for larger arrays/string, but since utf-8 is so common it may not make sense to avoid it.

spernsteiner: IMO the `BoundedSymbolic` impl for `String` should be able to produce any legal string, including UTF-8.  If you want ASCII-only for some test, you can do `String::bounded_symbolic_where::<N, _>("s", |s| s.is_ascii())`

podhrmic: @AlexMahrer can we add Stuart's suggestion into the comments please (about how to get ascii)?

RyanGlScott (commented):
Thanks, this is looking nice.

One notable feature that this PR does not currently implement is `derive` support for `BoundedSymbolic`, as described in parts (2) and (3) of https://github.com/GaloisInc/mir-json/issues/298. I'm OK with this limitation for now, although it would be worth opening a separate issue to track that.

podhrmic:
@AlexMahrer I guess that is a good point - being able to derive `BoundedSymbolic` would definitely be helpful.

AlexMahrer:
Added a derive macro [here](https://github.com/GaloisInc/mir-json/pull/318/changes#diff-666b68003c544f33f0983e20efdd1943aafad161d3200333da8b3ccfb013a08fR437). It could use some more tests.

AlexMahrer:
The flexibility of this derive macro is somewhat limited because all nested fields are bounded by the same constant passed to `BoundedSymbolic::bounded_symbolic::<N>`. Bounding the fields by different constants requires a custom impl of `BoundedSymbolic`. This becomes more painful the deeper in the structure the constants need to differ.

The same applies for any use of `bounded_symbolic_where`.

RyanGlScott:
My impression is that `BoundedSymbolic` is meant primarily as a convenience for quickly generating a value subject to some bound. It's always possible that you'll still need to hand-roll things in exceptional cases where the defaults aren't good enough. (As far as I am aware, this is a limitation that Kani's `BoundedArbitrary` also faces.)

libs/crucible_proc_macros/src/lib.rs
```diff
@@ -431,6 +434,252 @@ pub fn symbolic_derive(input: TokenStream) -> TokenStream {
     tokens.into()
 }
 
+/// Adds support for `#[cfg_attr(crux, derive(BoundedSymbolic))]`
+///
+/// This generates an implementation of the BoundedSymbolic trait
+/// that is completely symbolic. All fields will be generated using
```
RyanGlScott, line 440: Saying that it's "completely symbolic" is arguably misleading, given that the values are constrained to be bounded.

libs/crucible_proc_macros/src/lib.rs
```diff
@@ -431,6 +434,252 @@ pub fn symbolic_derive(input: TokenStream) -> TokenStream {
...
+/// their symbolic instances. In the case of enumerations, all
```
RyanGlScott, line 441: Saying "their symbolic instances" is now ambiguous, given that there are two different symbolic-oriented traits now (`Symbolic` and `BoundedSymbolic`). Given that you have a paragraph later that describes how fields are treated, perhaps it's best to leave off the "All fields will be generated using their symbolic instances" sentence.

libs/crucible_proc_macros/src/lib.rs
```diff
@@ -431,6 +434,252 @@ pub fn symbolic_derive(input: TokenStream) -> TokenStream {
...
+/// variants can be returned. Unions are not supported.
+///
+/// By default, the Symbolic trait will be implemented for each field unless
+/// that field is tagged with #[bounded] in which case the BoundedSymbolic
+/// trait will be implemented.
```
RyanGlScott, line 446: I don't think "the Symbolic trait will be implemented" is the right way to phrase this, since deriving `BoundedSymbolic` for a type will only implement it for that type alone, and not for any of the types of its fields. Perhaps a better way would phrase this would be something akin to "By default, all fields will be created using the `Symbolic` trait unless that field is tagged with `#[bounded]`, in which case the field will be created using the `BoundedSymbolic` trait."

libs/crucible_proc_macros/src/lib.rs
```diff
@@ -431,6 +434,252 @@ pub fn symbolic_derive(input: TokenStream) -> TokenStream {
...
+                        let field_inits = fields_named.named.iter().map(|field| {
+                            let fname = &field.ident;
+                            let ty = &field.ty;
+
+                            match has_bounded_attr(field) {
+                                Ok(false) => {
+                                    generics
+                                        .make_where_clause()
+                                        .predicates
+                                        .push(syn::parse_quote!(#ty: Symbolic));
```
RyanGlScott, line 562: There's something a bit odd about this in that it will stick a `where` clause for every field in a struct or enum, regardless of whether the field is a type parameter or not. This can lead to some funky generated code. For instance, if you have:

```rs
#[derive(BoundedSymbolic)]
struct NamedFields {
    field1: u32,
    field2: u64,
}
```

Then this will generate the following impl:

```rs
impl BoundedSymbolic for NamedFields where
    u32: Symbolic,
    u64: Symbolic,
{
    fn bounded_symbolic<const N : usize>(desc: &str) -> Self {
        Self {
            field1: <u32 as Symbolic>::symbolic(desc),
            field2: <u64 as Symbolic>::symbolic(desc),
        }
    }
}
```

It's technically not _wrong_ to do this (the code will still compile), but the `u32: Symbolic` and `u64: Symbolic` are redundant, and for larger data types, I imagine this will lead to pretty noticeable blowups in generated code size.

That being said, doing this in a smarter way will likely be non-trivial. Ideally, we'd only emit `Symbolic` or `BoundedSymbolic` constraints for type parameters, but figuring out which type parameters need which types of constraints requires some further analysis of the types involved. I think implementing this analysis is beyond the scope of this PR, so perhaps it would be best to leave a comment making note of this limitation for now.

(For what it's worth, Kani's `BoundedArbitrary` has a similar limitation, which they document [here](https://github.com/model-checking/kani/blob/e5f2e75a4a84e5b32e37aeae90b474ffc307af56/library/kani_macros/src/derive_bounded.rs#L15-L41).)

spernsteiner: Given [complexities around perfect derive](https://rkyv.org/derive-macro-features.html#omit_bounds), I think `derive(BoundedSymbolic)` should do non-perfect derive as in Kani's approach instead: emit an impl with `where T: Symbolic` (not `BoundedSymbolic`) for each type parameter, and document that `#[bounded]` can't be used on fields involving generics.  This would also be more consistent with `derive(Symbolic)`, which does non-perfect derive.

>figuring out which type parameters need which types of constraints requires some further analysis of the types involved

This analysis is not really feasible to implement in a proc macro.  Given `field: C<T>`, we might need `T: Trait`, or might not need it, or might need some other bound entirely (e.g. `Cell<T>: Clone` requires `T: Copy`), and figuring out which requires running the whole trait solver, which may in turn depend on the outputs of various proc macros (possibly even the proc macro we're currently evaluating).

RyanGlScott: TIL the term "perfect derive" (which, ironically, is far from perfect). I would be fine with using non-perfect drive as an alternative.

> This analysis is not really feasible to implement in a proc macro.

Agreed! I think I undersold how difficult this is by saying "some further analysis" :)

AlexMahrer:
Feedback on documentation and "perfect derive" issues has now been addressed.

RyanGlScott (approved):
Looks great. Thank you for your persistence on this!

libs/crucible_proc_macros/src/lib.rs
```diff
@@ -431,6 +434,191 @@ pub fn symbolic_derive(input: TokenStream) -> TokenStream {
...
+///
+/// This generates an implementation of the BoundedSymbolic trait that
+/// constructs symbolic values for all fields of the data type. In the
+/// case of enumerations, all variants can be returned. Unions are not
+/// supported.
+///
+/// By default, all fields will be constructed using the Symbolic trait
+/// unless that field is tagged with `#[bounded]` in which case the
+/// BoundedSymbolic trait will be used instead, bounding the size of
+/// the symbolic field by a specified constant.
```
spernsteiner: Which constant is used for the bound?  I assume the `const N` from the top-level `bounded_symbolic` call is passed down to each `#[bounded]` field?

libs/crucible_proc_macros/src/lib.rs
```diff
@@ -431,6 +434,191 @@ pub fn symbolic_derive(input: TokenStream) -> TokenStream {
...
+///
+/// while this results in a compile error:
+///
+/// ```rust
+/// #[derive(BoundedSymbolic)]
+/// struct Foo<T> {
+///     #[bounded]
+///     some_field: T,
+/// }
+/// ```
```
spernsteiner: This comment should note that the underlying issue is that the derived impl will have `where T: Symbolic` for each type parameter T, rather than `T: BoundedSymbolic` or a perfect-derive constraint like `Vec<T>: BoundedSymbolic`.  That way, if the user goes beyond simple cases like `#[bounded] field: Vec<T>` vs `#[bounded] field: T`, they have enough information to figure out how this restriction applies to their types.

spernsteiner (approved):

